PR #15, opened 2016-08-24
Update README.md

### LINKED ISSUE #14: Readme missing the option for installing nodejs-legacy
After installing nodejs, we also need to install nodejs legacy ```sudo apt install nodejs-legacy```
Without this, `npm install` throws errors.

the-dagger:
Fixes #14 
@niranjan94 Please check and see if it is necessary or not.


niranjan94:
See comment on #14

PR #38, opened 2017-03-06
Two alert dialogs issue solved

### LINKED ISSUE #37: Two alert dialogs appearing if no network connection
On starting app without network connection, two alert dialogs are created saying `No Internet Connection`.

mejariamol:
fixes #37

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -21,6 +21,7 @@ export class OrganizerAppComponent {
 
   @ViewChild(Nav) public nav: Nav;
   public rootPage: any;
+  private disconnectSubscription: any;
 
   constructor(private platform: Platform, private queueService: QueueService,
               private networkCheckService: NetworkCheck, private menuCtrl: MenuController) {
@@ -50,13 +51,19 @@ export class OrganizerAppComponent {
   }
 
   private handleNetworkChanges() {
-    Network.onDisconnect().subscribe(() => {
-      this.networkCheckService.showNoNetworkAlert();
-    });
+    this.initDisconnectSubscription();
     Network.onConnect().subscribe(() => {
+      this.initDisconnectSubscription();
       setTimeout(() => {
         this.queueService.processQueue();
       }, 3000);
+    });
+  }
+
+  private initDisconnectSubscription() {
+    this.disconnectSubscription = Network.onDisconnect().subscribe(() => {
+      this.networkCheckService.showNoNetworkAlert();
+      this.disconnectSubscription.unsubscribe();
     });
   }
 }


src/app/app.component.ts
```diff
@@ -21,6 +21,7 @@ export class OrganizerAppComponent {
 
   @ViewChild(Nav) public nav: Nav;
   public rootPage: any;
+  private disconnectSubscription: any;
```
niranjan94: Since we know the type, don't use `any`

src/app/app.component.ts
```diff
@@ -50,10 +51,15 @@ export class OrganizerAppComponent {
   }
 
   private handleNetworkChanges() {
-    Network.onDisconnect().subscribe(() => {
+    this.disconnectSubscription = Network.onDisconnect().subscribe(() => {
```
niranjan94: We have code-duplication here. Please avoid duplication of code.

src/app/app.component.ts
mejariamol: I imported `Subscription` using `import {Subscription} from '@reactivex/rxjs/es6/Subscription.js'`.
Still giving error `Cannot find name 'Subscription'`.
Is there anywhere else where I will need to import the package??

niranjan94: @mejariamol just a simple `import {Subscription} from "rxjs";` should do ...

mejariamol: using `import {Subscription} from "rxjs";` 
giving error ` Type 'Subscription' is not assignable to type 'Subscription'. Property '_parent' is missing in type
            'Subscription'.`
what can be the problem?

niranjan94: Do a `typeof` and see what is the actual type of `disconnectSubscription`

mejariamol: it is returning `object`

niranjan94: Weird. :thinking: Okay .. use `any` for now .. I'll check this out tomorrow

src/app/app.component.ts
niranjan94: Fix this one. And it's ready to merge

mejariamol:
@niranjan94 The function `initDisconnectSubscription()` itself unsubscribes the subscription once error is shown. Hence there won't be any case of subscribing to `onDisconnect` twice.

niranjan94:
@mejariamol one PR per issue please. Don't mix code/

mejariamol:
@niranjan94 Shall I remove last commit?

niranjan94:
Yes @mejariamol ... Remove and open a new PR

mejariamol:
@niranjan94 Removed last commit. To make new PR for #39 this will need to be merged before right?

niranjan94:
@mejariamol I don't see why this should be merged before making another PR ... You can always create a new branch and make a new PR

PR #61, opened 2017-03-13
Implemented pull to refresh feature

### LINKED ISSUE #41: Implement pull to refresh on all pages containing data load from server
Presently there is no page refreshing feature on the pages which display data from the server.
The pages are
- `EventAttendeesPage`
- `EventDashboardPage`
- `EventsPage`

`SwipeRefreshLayout` from android like feature is preferred.

mejariamol:
fixes #41

### DIFF
--- a/src/pages/event-attendees/event-attendees.ts
+++ b/src/pages/event-attendees/event-attendees.ts
@@ -5,7 +5,8 @@ import {BarcodeScanner} from "ionic-native";
 import {IAttendee} from "../../interfaces/attende";
 import {IEvent} from "../../interfaces/event";
 import {AttendeesService} from "../../services/attendees.service";
-import {QueueService} from "../../services/queue.service";
+import { QueueService } from "../../services/queue.service";
+import { NetworkCheck } from "../../services/network-check.service";
 
 @Component({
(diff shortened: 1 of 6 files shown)


src/pages/event-attendees/event-attendees.ts
```diff
@@ -56,6 +57,25 @@ export class EventAttendeesPage {
     });
   }
 
+  public doRefresh(refresher) {
+    this.attendeesService.loadAttendees(this.event.id).subscribe(
```
niranjan94: Let's not have duplication. Please do it in such that there is no duplication.

mejariamol: @niranjan94 `refresher.complete()` needs to be called after the async work. Thats why I couldn't make a common function. Please suggest how can be duplication avoided in this case?

niranjan94: You can pass a custom callback and execute it

mejariamol: @niranjan94 Avoided duplication. But not using callback. Please check if the approach is fine.

PR #96, opened 2017-05-15
feat: Refactor to MVP and implement login

### LINKED ISSUE #93: Implement Login MVP
Part solving #92 - MVP Implementation

- Use Retrofit instead of Volley ( Easy of Testing )
- Use RxJava for Observable pattern rather than callbacks ( easy testing with pluggable schedulers and no need of callback captures), with additional feature of offloading tasks over Main Thread
- Use Retrolambda to reduce anonymous inner classes and use lambda expressions itself

Proposed Architecture: 

Package Style : 
- contract (Containing interfaces)
    - model ( Contains Repository pattern model too )
    - view
    - presenter
- data
    - network ( Contains Retrofit implementation )
    - Model Implementation
- ui
    - presenter
        - Login Presenter Implementation
    - activities
        - LoginActivity ( View Implementation )

Any suggestions are welcome

iamareebjamal:
Fixes #93 

- Implement MVP architecture
- Add LoginPresenter, LoginModel and UtilModel and their unit tests


iamareebjamal:
This is the last update with so many changes as this was changing the package architecture of the app to MVP compliant

Only actual change is addition of retrofit, rxjava, retrolambda and MVP implementation of Login Screen

I have tested it myself

Please test @Ronak-59 @mejariamol @Rishabhk07 so we can merge

app/src/main/java/org/fossasia/openevent/app/contract/view/LoginView.java
```diff
@@ -0,0 +1,11 @@
+package org.fossasia.openevent.app.contract.view;
+
+public interface LoginView {
+
+    void showProgressBar(boolean show);
```
mejariamol: @iamareebjamal I think `showProgressBar` will not be required by Presenter. It can be called by the view itself.

app/src/main/java/org/fossasia/openevent/app/contract/presenter/LoginPresenter.java
```diff
@@ -0,0 +1,15 @@
+package org.fossasia.openevent.app.contract.presenter;
+
+import org.fossasia.openevent.app.contract.view.LoginView;
+
+public interface LoginPresenter {
+
+    void start();
```
mejariamol: @iamareebjamal What are these methods for `start` and `stop` ?

app/src/main/java/org/fossasia/openevent/app/contract/presenter/LoginPresenter.java
```diff
@@ -0,0 +1,15 @@
...
+
+    void stop();
+
+    void login(String email, String password);
+
+    LoginView getView();
```
mejariamol: @iamareebjamal This is not required. as we can pass the view to presenter through the constructor from the view itself.

iamareebjamal: This is for testing

app/src/main/java/org/fossasia/openevent/app/contract/presenter/LoginPresenter.java
iamareebjamal: Analogous to attaching or detaching view lifecycle but more general for any starting activity and stopping activity to do cleanup and release view to avoid memory leaks

app/src/main/java/org/fossasia/openevent/app/contract/view/LoginView.java
iamareebjamal: It should be implemented by View obviously as it will be called by presenter when need of doing an asynchronous task

iamareebjamal: Please see the `View` interface here

https://medium.com/@cervonefrancesco/model-view-presenter-android-guidelines-94970b430ddf

mejariamol: I am saying progressbar **can be called by the view itself** when it will call presenter's login method and can be dismissed in the onSuccess or onError functions. Hence **will not require to be called by presenter**.
Hence need not to be **implemented** in the view.

app/src/main/java/org/fossasia/openevent/app/contract/presenter/LoginPresenter.java
mejariamol: @iamareebjamal Just to get clear - who will be calling these functions ??  View?

app/src/main/java/org/fossasia/openevent/app/contract/view/LoginView.java
iamareebjamal: View should be dumb, not knowing what the logic of doing anything should be. What you are saying is the MVC approach, we don't want that. We want presenter to decide what to do with view, view should just do what it is told. showData means show data, not stop loading, save in cache, update model. Everything requiring condition and login must be done in model or presenter

The onError and onSuccess are too generic, it should rather be, showError and showEventActivity because view should not know what success or error mean. Besides if we let view decide that, how we test that view actually calls it? That was the sole reason we were moving logic away from view to abstractions

Also, showError doesn't necessarily mean the loading was on. Suppose that we verify email format before sending it to server, we will still show error but not show the progressbar, this should thus be handled by presenter

app/src/main/java/org/fossasia/openevent/app/contract/presenter/LoginPresenter.java
iamareebjamal: Correct. View will be notifying the presenter that it is now ready to attach or detach

iamareebjamal:
@niranjan94 I think we should merge as the changes are trivial and I have also written thorough tests for both LoginPresenter and LoginModel and both are passing

Also, I have tested the login under various conditions and network errors, I'd say it's more robust and covered than before

And sorry for all for this large a PR, needed to refactor the structure to MVP

Issue #92, opened 2017-05-12
MVP - Model View Presenter pattern Implementation


mejariamol:
Currently the project is based on MVC (Model View Controller) pattern in which unit testing is difficult. Hence we will be implementing MVP pattern which facilitates unit testing.

Expected: Design MVP interface containing the required interfaces.

iamareebjamal:
I have my exams now so I may only be helping in the design of the architecture :)

mejariamol:
@iamareebjamal @Ronak-59 @Rishabhk07 We should consider using [Mosby MVP Library](https://github.com/sockeqwe/mosby) for implementing MVP. It will make the process relatively easier than starting from the scratch.

iamareebjamal:
@mejariamol I think it ties our hand to a particular convention we might not need to follow

Furthermore, I have created a basic structure for app and we might also use dagger 2 in future in accordance to our needs

iamareebjamal:
Also, I think it is basic enough so that we learn to do from scratch rather than relying on an opinionated platform so that we can adapt it according to our needs, like breaking the model temporarily for rapid prototyping

PR #102, opened 2017-05-16
feat: Implement MVP for Event

iamareebjamal:
Fixes part of #92 

- Add MVP for Event
- Add thorough tests
- Add logout feature


iamareebjamal:
Some tests remain to add

app/src/main/java/org/fossasia/openevent/app/data/cache/ObjectCache.java
```diff
@@ -0,0 +1,34 @@
+package org.fossasia.openevent.app.data.cache;
+
+
+import android.support.v4.util.LruCache;
+
+import org.fossasia.openevent.app.contract.model.CacheModel;
+
+public class ObjectCache implements CacheModel {
+
+    private static ObjectCache instance;
```
mejariamol: @iamareebjamal we should [implement singleton with enum](http://stackoverflow.com/questions/26285520/implementing-singleton-with-an-enum-in-java). Current pattern  will not support unit testing.

app/src/main/java/org/fossasia/openevent/app/contract/model/UtilModel.java
```diff
@@ -2,10 +2,20 @@
 
 public interface UtilModel {
```
mejariamol: @iamareebjamal No need to make interface UtilModel.. Just define static methods in AndroidUtilModel class. And prefer `Util.java` for class name.

app/src/main/java/org/fossasia/openevent/app/data/cache/ObjectCache.java
iamareebjamal: You can see the test I have written for ObjectCache, It works

app/src/main/java/org/fossasia/openevent/app/contract/model/UtilModel.java
iamareebjamal: We need to mock interactions, there is no better way to do so than to create interface of it. And besides, it is always better to program in abstractions

iamareebjamal: Secondly, UtilModel classes depend on context and thus must be abstracted. Holding context in a static variable is a serious memory leak

app/src/main/java/org/fossasia/openevent/app/data/cache/ObjectCache.java
iamareebjamal: And as it is not dependent on android, we need not mock its behaviour, we can just use actual object and see if it works

iamareebjamal:
@niranjan94 It is completed. Please merge if check passes :smile:

app/src/main/java/org/fossasia/openevent/app/contract/model/EventModel.java
```diff
@@ -0,0 +1,18 @@
+package org.fossasia.openevent.app.contract.model;
+
+import org.fossasia.openevent.app.data.models.Event;
+import org.fossasia.openevent.app.data.models.User;
+
+import java.util.List;
+
+import io.reactivex.Observable;
+
+public interface EventModel {
```
mejariamol: @iamareebjamal please make changes in the interfaces(created in this commit) names as we discussed earlier.

mejariamol: No problem. I will make these changes in the new PR.

iamareebjamal: Actually, your PR should handle this because it is not the issue of this PR

PR #103, opened 2017-05-17
feat: Add butterknife to remaining activity

iamareebjamal:
Add butterknife to remaining LoginActivity


iamareebjamal:
@niranjan94 Please review

PR #109, opened 2017-05-19
test: Add more feature and test in Cache model

iamareebjamal:
- Add `getSize` in cache
- Add more tests for size and duplicate check


iamareebjamal:
@niranjan94 Please merge

PR #111, opened 2017-05-19
feat: Convert Event Detail to MVP

iamareebjamal:
- Modify the contracts accordingly to accommodate new information
- Add M, V, and P for event details
- Generalized Event Model with parameters
- Modify UI to generalize
- Add tests for attendees


iamareebjamal:
screenshot

app/src/main/java/org/fossasia/openevent/app/contract/model/IEventRepository.java
```diff
@@ -1,18 +1,22 @@
...
+import org.fossasia.openevent.app.data.models.Attendee;
 import org.fossasia.openevent.app.data.models.Event;
 import org.fossasia.openevent.app.data.models.User;
 
 import java.util.List;
 
 import io.reactivex.Observable;
 
-public interface IEventModel {
+public interface IEventRepository {
```
mejariamol: @iamareebjamal name interface as `EventDataSource` and let `EventDataRepository` implement it. `*DataSource` - `*DataRepository` (`*Source` for interface and `*Repository` for the class implementing it) naming convention is better. Also apply this to other models.

app/src/main/java/org/fossasia/openevent/app/data/network/api/RetrofitEventRepository.java
```diff
@@ -0,0 +1,87 @@
...
+import org.fossasia.openevent.app.data.models.User;
+import org.fossasia.openevent.app.utils.Constants;
+
+import java.util.List;
+
+import io.reactivex.Observable;
+import io.reactivex.android.schedulers.AndroidSchedulers;
+import io.reactivex.schedulers.Schedulers;
+
+public class RetrofitEventRepository implements IEventRepository {
```
mejariamol: @iamareebjamal same here. `EventDataRepository` would be enough. Just name the related interface as `EventDataSource`.

app/src/main/java/org/fossasia/openevent/app/contract/model/IEventRepository.java
iamareebjamal: Actually Repository is an interface, it is a model which abstracts the pattern of loading data from multiple sources, namely disk, cache or network

However, your second point is very correct. It should be just EventDataRepository

Why we want to have same name is that DataSource != DataRepository

DataSource is like NetworkDataSource DiskDataSource, so when we create a persistent model, it will class with the repository interface

mejariamol: `EventDataRepository` suggests something like repository which contains data and can be retrieved directly from it. Hence it can be the class name implementing the model interface. 
For interface, it can be called as `DataSource`.
It is just a suggestion. :) I thought it in this way.

iamareebjamal: Yeah that was what I was trying to tell. A repository may contain several data sources, disk cache and network, so that's why we can't name its interface to be DataSource

iamareebjamal:
@niranjan94 Please merge this along with the previous two PRs

PR #115, opened 2017-05-22
refactor: Organise tests into corresponding packages

iamareebjamal:
- Create subpackages for presenter, models and utility tests
- Delete duplicate model and test


iamareebjamal:
@niranjan94 Please merge

PR #116, opened 2017-05-22
implemented MVP patten on attendees page

### LINKED ISSUE #114: Implement MVP on attendees page
Sub issue #92 

Expected: Implement MVP on attendees page and add related tests.

I am working on this.

mejariamol:
fixes #114


app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesActivity.java
```diff
@@ -29,22 +32,25 @@
...
+    ProgressBar progressBar;
+
     static ArrayList<Attendee> attendeeArrayList = new ArrayList<>();
     AttendeeListAdapter attendeeListAdapter;
 
     long id;
     public static final int REQ_CODE = 123;
     public static final String TAG = "AttendeeListActivity";
 
+    IAttendeesPresenter attendeesPresenter;
```
iamareebjamal: Please use package level access `private`

app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesActivity.java
```diff
@@ -59,36 +65,20 @@ protected void onCreate(Bundle savedInstanceState) {
...
-                Intent i = new Intent(AttendeesActivity.this, ScanQRActivity.class);
-                startActivityForResult(i, 123);
-            });
-        } else {
-            Toast.makeText(this, Constants.NO_NETWORK, Toast.LENGTH_SHORT).show();
-        }
+    @Override
+    protected void onDestroy() {
+        super.onDestroy();
+        attendeesPresenter.dettach();
```
iamareebjamal: Correct the spelling to `detach`

app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesActivity.java
```diff
@@ -121,34 +111,39 @@ public void checkInAlertBuilder(final int index) {
             + "Ticket: " + thisAttendee.getTicket().getType())
             .setPositiveButton("OK", (dialog, which) -> {
                 Log.d(TAG, "onClick: inside ok");
-                changeCheckStatus(thisAttendee.getId(), index);
+                attendeesPresenter.toggleAttendeeCheckStatus(index, thisAttendee.getId());
```
iamareebjamal: Why passing the index? Pass the attendee to operate on instead

Index is extremely unreliable and will be difficult to track bugs when we implement sorting and filtering in future

app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesActivity.java
```diff
@@ -121,34 +111,39 @@ public void checkInAlertBuilder(final int index) {
...
-                @Override
-                public void onError(VolleyError error) {
-
-                }
-            });
-        } else {
-            Toast.makeText(this, Constants.NO_NETWORK, Toast.LENGTH_SHORT).show();
+    @Override
+    public void showProgressBar(boolean show) {
+        if(show) {
```
iamareebjamal: Use Utils function `showView`

app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesActivity.java
```diff
@@ -121,34 +111,39 @@ public void checkInAlertBuilder(final int index) {
...
+    public void loadAttendeesIntoList(List<Attendee> attendees) {
+        attendeeArrayList.clear();
+        attendeeArrayList.addAll(attendees);
+        attendeeListAdapter.notifyDataSetChanged();
+    }
 
+    @Override
+    public void updateAttendee(int position, Attendee attendee) {
+        attendeeArrayList.set(position, attendee);
+        attendeeListAdapter.notifyDataSetChanged();
```
iamareebjamal: Don't use index, because it will not remain constant till the callback comes back.

For instance, if we implement sorting in future, and user sorts using different criteria before the response comes back, then this code will update the wrong attendee

app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesPresenter.java
```diff
@@ -0,0 +1,68 @@
...
+import org.fossasia.openevent.app.event.attendees.contract.IAttendeesPresenter;
+import org.fossasia.openevent.app.event.attendees.contract.IAttendeesView;
+
+public class AttendeesPresenter implements IAttendeesPresenter {
+
+    IEventDataRepository eventDataRepository;
+    IAttendeesView attendeesView;
+    Long eventId;
+
+    public AttendeesPresenter(IAttendeesView attendeesView, Context context, Long eventId) {
```
iamareebjamal: Never give `Context` or any Android specific object to presenter, use `UtilModel` instead, it is specifically made to handle context based actions

app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesPresenter.java
```diff
@@ -0,0 +1,68 @@
...
+        this.attendeesView = attendeesView;
+        eventDataRepository = new EventDataRepository(new UtilModel(context));
+        this.eventId = eventId;
+    }
+
+    @Override
+    public void attach() {
+        loadAttendees(true);
```
iamareebjamal: Why setting `forceReload` parameter to true? This should mean that we want to force the results from server, and not use cached results, which we don't want by default

app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesPresenter.java
```diff
@@ -0,0 +1,68 @@
...
+    public void dettach() {
+        attendeesView = null;
+    }
+
+    @Override
+    public void loadAttendees(boolean forceReload) {
+        attendeesView.showProgressBar(true);
+        eventDataRepository.getAttendees(eventId, forceReload)
+            .subscribe(attendees -> {
+                attendeesView.loadAttendeesIntoList(attendees);
```
iamareebjamal: Better naming here would be just `showAttendees` 

Be declarative and avoid specific command like `loadIntoList`

app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesPresenter.java
```diff
@@ -0,0 +1,68 @@
...
+                attendeesView.showProgressBar(false);
+            }, throwable -> {
+                attendeesView.showProgressBar(false);
```
iamareebjamal: Maintain a strict pattern of commands so that tests are very compatible with each other,

If you disable progressbar after results in success, you should disable it after error too

app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesPresenter.java
```diff
@@ -0,0 +1,68 @@
...
+                attendeesView.showErrorMessage(throwable.getMessage());
+            });
+    }
+
+    @Override
+    public void loadAttendees() {
```
iamareebjamal: Why 2 methods? Only loadAttendees(boolean)` would be enough

app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesPresenter.java
```diff
@@ -0,0 +1,68 @@
...
+        loadAttendees(false);
+    }
+
+    @Override
+    public void toggleAttendeeCheckStatus(int position, Long attendeeId) {
+        eventDataRepository.toggleAttendeeCheckStatus(position, eventId, attendeeId)
```
iamareebjamal: Show progressbar till the action is completed

app/src/main/java/org/fossasia/openevent/app/event/attendees/contract/IAttendeesPresenter.java
```diff
@@ -0,0 +1,17 @@
...
+
+    void dettach();
+
+    void loadAttendees(boolean forceReload);
+
+    void loadAttendees();
+
+    void toggleAttendeeCheckStatus(int position, Long attendeeId);
+
+    IAttendeesView getView();
```
iamareebjamal: Presenter should not declare `getView` as view does not need to call it. Add it as a public method in presenter implementation because test needs it, but not in the contract itself

app/src/main/java/org/fossasia/openevent/app/data/EventDataRepository.java
```diff
@@ -83,6 +83,18 @@ public EventDataRepository(IUtilModel utilModel) {
         return getData(() -> eventService.getEvents(authorization), EVENTS, reload);
     }
 
+    @Override
+    public Observable<Attendee> toggleAttendeeCheckStatus(long position, long eventId, long attendeeId) {
+        return eventService.toggleAttendeeCheckStatus(eventId, attendeeId, authorization)
+            .doOnNext(attendee -> {
+                List<Attendee> attendees = (List<Attendee>) cacheModel.getValue(ATTENDEES + eventId);
```
iamareebjamal: You can't rely, that on return of this function, cache will give you a non null attendee list.

What if it was never saved or cleared due to some reason like force reload?

app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesActivity.java
mejariamol: index will be used to update the attendee in the attendees list once the check in state is toggled.

app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesPresenter.java
mejariamol: >and not use cached results, which we don't want by default

why don't we want the results from the server? Cache should be preferred on no network only. What do you say?

app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesPresenter.java
mejariamol: `showAttendees` seems like hiding and showing the attendees list view. But actually the method is populating attendees into the list

app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesActivity.java
iamareebjamal: It's an unreliable way to do it. Please read my argument

app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesPresenter.java
iamareebjamal: Exactly. So forceReload must be set to false

iamareebjamal: See other presenters

app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesPresenter.java
iamareebjamal: View doesn't know what is list, view should just be provided with just the data and said to show it. We are definitely hiding and showing attendees, because in case there are no attendees, we will hide attendees and show empty view in future

app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesActivity.java
iamareebjamal: Index of that item in the list may change till response is returned

mejariamol: response will be returned in less time. hence very less chance of list change in that small period which is acceptable. 
Any other way to update the list after toggling?

app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesPresenter.java
mejariamol: `forceReload` set to `false` means use cached data right? that is why I am setting it to `true` so that it won't use cached data when network is available. Cached data will be used only when network is not available.

app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesActivity.java
iamareebjamal: Actually, you can't assume that network is reliable and simple sorting will definitely change the index .

Better would be to, just send the attendee to get updated and once it comes back updates, presenter searches the local list for the item with the same ID, and sends the updated list to view

This is definitely a little slower than previous method but is clean, reliable and has no dependency on the data managed by view

iamareebjamal: Additionally, we are doing everything asynchronously so there will be no performance hit

iamareebjamal: This method also has a very small complexity and is very trivial to implement using RxJava `filter` and `firstOrError`

app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesPresenter.java
iamareebjamal: Actually, we need to show data as fast as possible. If it is stale, the user can trigger an update, or we can automatically do it, but we need to show the data which is persisted. That's the point of saving it. Showing it offline only would be mundane as user already knows it is stale

app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesPresenter.java
```diff
@@ -0,0 +1,66 @@
...
+
+public class AttendeesPresenter implements IAttendeesPresenter {
+
+    private IEventDataRepository eventDataRepository;
+    private IAttendeesView attendeesView;
+    private long eventId;
+
+    public AttendeesPresenter(IAttendeesView attendeesView, Context context, Long eventId) {
+        this.attendeesView = attendeesView;
+        eventDataRepository = new EventDataRepository(new UtilModel(context));
```
iamareebjamal: Add EventDataRepository in constructor. Presenter should not get hold of context and not resolve any dependency itself

iamareebjamal: We'll be probably using DI for injecting dependencies once this last component of MVP is added and tested so we won't have to deal with all this hassle. Till then, pass it in constructor after creating in the Activity

iamareebjamal: Also, use primitive `long` for Event ID, prefer primitives over objects

app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesPresenter.java
iamareebjamal: Just imagine how much user will have to wait for every attendee information? This is an anti-pattern to always fetch from the internet. Suppose, there are 10,000 attendees. How much time it will take to download that info if we don't use the cached versions

mejariamol: first of all we will not be loading 10,000 attendees at once. probably with the next-gen api we will be getting the attendees with limit. And for the better user experience the approach shall be initially display the cached data and fetch updated data from the server in presence of network.
hence we can go for something like
```
def attach() {
    loadAttendees(false);
    loadAttendees(true);
}

app/src/main/java/org/fossasia/openevent/app/data/EventDataRepository.java
mejariamol: btw we are going to use some local database. hence for now I will just put the condition for non nullity of the cached list.

mejariamol:
@iamareebjamal please write unit tests for `AttendeesPresenter` for now. I will learn to write tests shortly.

app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesPresenter.java
iamareebjamal: The essence of MVP is never to delegate a behavior of one class to another class. If we want to define the behavior of model such that it updates the result every time we fetch, then we will define it in model and presenter will just call for data and it will return the cached result and then update the view when the view when data is available automatically, but till it is required, let's take it literally, load from cache if it's saved, and use network only if it's not. Since we are using in-memory cache, the results will never persist across sessions, so we never force user to see too stale data.

So, *when* the model is coded to be smart enough, it will be triggered to give us cached result and the model itself will fetch the result and save it to refresh the view later. 

Read about offline first apps, where we act locally and sync globally. We want to make the experience as smooth as possible, but we can't jump to conclusions about what users need. We must assume they don't want to fetch the data every second and if they are offline, the current code won't show the previous data. So, we must always at least show the cached data, doesn't matter if we fetch 10000 results or not

iamareebjamal:
Sure, I will write the tests today and push them

iamareebjamal:
Will add tests soon

iamareebjamal:
PR is getting large. Let's handle tests in another one. Basic tests have been added nevertheless

PR #120, opened 2017-05-24
feat: Share event parcel to display initial info

### LINKED ISSUE #119: Pass loaded event data to Event Detail
Currently, the Events activity passes the event id to event detail activity which loads the event from cache or network.

The basic info can be shown while the ticket info is being loaded by passing the event parcel from events activity to event detail activity

I'm working on it

iamareebjamal:
Fixes #119 

- Make Event Parcelable
- Share event to EventDetailActivity to show initial information while ticket info loads

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/app/data/models/Event.java
+++ b/app/src/main/java/org/fossasia/openevent/app/data/models/Event.java
@@ -1,10 +1,13 @@
 package org.fossasia.openevent.app.data.models;
 
+import android.os.Parcel;
+import android.os.Parcelable;
+
 import com.google.gson.annotations.SerializedName;
 
 import java.util.List;
 
-public class Event {
+public class Event implements Parcelable {
 
     @SerializedName("background_image")
     private String backgroundImage;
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/fossasia/openevent/app/event/detail/EventDetailsActivity.java
```diff
@@ -50,20 +54,24 @@ protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         setContentView(R.layout.activity_event_details);
 
-        ButterKnife.bind(this);
+        Event initialEvent = getIntent().getParcelableExtra(EVENT_KEY);
+
+        if(initialEvent == null) {
```
mejariamol: what will be the case for this to happen? and `finish` is not good option. It has `event_id`, hence can fetch event from `eventsDataRepository` to show required info.

app/src/main/java/org/fossasia/openevent/app/data/models/Event.java
```diff
@@ -366,15 +369,85 @@ public void setVersion(Version version) {
     @Override
     public String toString() {
```
iamareebjamal: When we log for debug, we want to see the contents of models

app/src/main/java/org/fossasia/openevent/app/data/models/Event.java
```diff
@@ -366,15 +369,85 @@ public void setVersion(Version version) {
...
+
+    // Parcelable Information - Only basic event info
+
+    @Override
+    public int describeContents() {
+        return 0;
+    }
+
+    @Override
+    public void writeToParcel(Parcel dest, int flags) {
```
iamareebjamal: I checked the time to parcel - 1 ms

Changing parcel again and again will require extra overhead. I have only selected non nested string fields and tested for the speed

app/src/main/java/org/fossasia/openevent/app/event/detail/EventDetailsActivity.java
iamareebjamal: It does not have id. The case will be when programmer forgets to pass the event to the activity. We can't trust anyone. Showing a blank screen is worse than crashing the app. At least second case tells you that something is wrong.

If I forget to add event as intent extra, it will give NPE, so I have to check if it is indeed passed and log and exit if it's not

iamareebjamal: Same thing with id. First, the code was assuming that it has been given the event id and not even checking its validity and the default id was 0. There is hopefully no event with id 0, or else it would have shown its info if id wasn't passed and a whole array of authentication errors which would have made the user confused.

And even if there's no event with ID 0, it shouldn't have executed normally

mejariamol: > The case will be when programmer forgets to pass the event to the activity. We can't trust anyone.

So this is for development stage right?

>Showing a blank screen is worse than crashing the app.

At developing stage crashing the app will be better option as it will suggest something wrong. No need to adjust that case.

iamareebjamal: No man, it is not for development stage, it's for production - for null safety

iamareebjamal: Tell me, how can you assume that the event object was passed to you (even if it was) ? You have to check if it was actually past. If you are telling me to remove null check because we are definitely passing the event currently, remember that this is open source project. If there comes a point where someone forgets to pass a parameter to event detail activity, would you want the app to crash?

Always check nullable values - it's this strategy, not anything else

mejariamol: >If there comes a point where someone forgets to pass a parameter to event detail activity, would you want the app to crash?

yes, so that it can be fixed. Not that important, I just thought adjusting that case will make hard for developer to notice that bug.

iamareebjamal: It's actually opposite, and we are writing tests anyway

iamareebjamal:
@niranjan94 Please merge

PR #123, opened 2017-05-25
feat: Implement MVP for QR Scan Activity

### LINKED ISSUE #92: MVP - Model View Presenter pattern Implementation

iamareebjamal:
Fixes #92 

- Add protocol for communication in QR Activity and presenter
- Make efficient use of concurrency to load camera and bar code scanner modules
- `Debounce` and consume only appropriate results by `distinctUntilChanged` 
- Show progressbar while loading camera
- Show current scan code to user
- Show conversion to green if accepted
- Refactor QR to its own package

photo6143426851633211307


iamareebjamal:
This was fun to write and captures the full essence of MVP. Since it'll be hard to test because of several layers of interconnected communication, will add tests tomorrow in separate PR

iamareebjamal:
@mejariamol Please review

app/src/main/java/org/fossasia/openevent/app/qrscan/ScanQRPresenter.java
```diff
@@ -0,0 +1,116 @@
...
+            .subscribeOn(Schedulers.io())
+            .observeOn(AndroidSchedulers.mainThread())
+            .subscribe(this::processBarcode);
+    }
+
+    private void processBarcode(String barcode) {
+        scanQRView.showData(barcode);
+
+        Observable.fromIterable(attendees)
+            .filter(attendee -> (attendee.getOrder().getIdentifier() + "-" + attendee.getId()).equals(barcode))
```
mejariamol: we can shift this checking to `eventDataRepository`, so no need to maintain attendees list in presenter itself. 

1. presenter will pass the scanned identifier to `eventDataRepository` 

2. accordingly `eventDataRepository` will check the validity and return the event.

iamareebjamal: Data repositories work is only one to provide data in efficient manner from different sources, and to process data only to facilitate efficient transfer of it throughout the app. We don't want to make a God class doing everything related to events

By this argument, no class should hold attendee, event detail and for processing, just ask event repo -> Not true, we are modularizing, one class does only one thing and does it good. This is not the work of event repository to check barcodes or calculate checked in attendees per total

app/src/main/java/org/fossasia/openevent/app/qrscan/ScanQRActivity.java
```diff
@@ -0,0 +1,244 @@
...
+        AndroidUtils.setTint(barcodePanel, ContextCompat.getColor(this, R.color.green_a400));
+
+        Intent resultIntent = new Intent();
+        resultIntent.putExtra(Constants.SCANNED_ATTENDEE, attendee);
+        setResult(AttendeesActivity.REQ_CODE, resultIntent);
+        finish();
+    }
+
+    @Override
+    public void showInView(boolean show) {
```
iamareebjamal: Yeah sorry this one must have skipped my mind

app/src/main/java/org/fossasia/openevent/app/qrscan/ScanQRPresenter.java
mejariamol: >By this argument, no class should hold attendee

the classes which need the list should hold the list. like `AttendeesActivity`.

I think scanned identifier validation like functionalities should be moved to `eventDataRepositiory` as then these can be called from anywhere in the app wherever required. 

Let's take @niranjan94 view?

iamareebjamal: @mejariamol This is not the task of EventRepository, if we really need to refactor it, it should be in another model. 

And which other class except QR Scan Presenter may need to check identifier against barcode? :astonished:

iamareebjamal: This can be done -> Create `AttendeeModel`, use it to 
- Toggle Status Check
- Match against barcode (Not really sure though why we need it out of QR Presenter)

If we do this, we move out toggleAttendee from EventRepository because that too isn't the job of a repository

mejariamol: But I am thinking if we bring all the data related functionality to a single class it will be handy to pass that single class object to any presenter which will handle all the data logic.

iamareebjamal: That is exactly opposite of what we are trying to do. Good Projects have many dependencies of many small components and a dependency injector manages them. Ideally, a class should not have more than 500 lines and a single core functionality. 

Look at `RxJava` project, there are so many components and only 2-3 classes are coupled to each other. In modular design, w don't group functionality based on the frequency of usage so that we can pass a single object through the code, rather than that we create several components and let a DI distribute them throughout.

The next step after completion of this PR will be to add Dagger 2

mejariamol: cool. :+1:

iamareebjamal:
@niranjan94 Please merge

iamareebjamal:
Resolved conflicts @niranjan94

PR #126, opened 2017-05-26
chore: Enable proguard and add release app to Travis

iamareebjamal:
Fixes #126 

- Enable and configure Proguard
- Add release apk to Travis


iamareebjamal:
@niranjan94 Please review

iamareebjamal:
@niranjan94 There was variable resolution error in the script. It resolved to full path rather than relative and thus couldn't copy the apk

Shall I revert and fix?

Also, how to check travis on PRs that it will work correctly?

Sorry for this, I didn't see it coming

niranjan94:
@iamareebjamal yes. Please fix it. 
You can temporarily make a commit + PR with the PR check commented out to see if it works.

PR #136, opened 2017-05-27
fix: Travis build and test apk upload

iamareebjamal:
Fix travis upload of built apks


iamareebjamal:
@niranjan94 Fixed. Please review

PR #142, opened 2017-05-29
Add Master/Detail Flow support to the event list

### LINKED ISSUE #128: Use Fragments instead of Activities
- Fragments are lightweight
- Communication between fragments doesn't need parceling/un parceling
- We can retain fragment state across configuration changes
- Easily manage states and transitions using backstack
- Can be used to create Tablet Layout

@mejariamol I have consecutive exams for a few days. Can you take this up?

### LINKED ISSUE #141: Use support library's Toolbar to implement action bar
To be done with #128 

Currently the activities use default app theme which has Actionbar as an app bar.

We should use the support library's [Toolbar](https://developer.android.com/reference/android/support/v7/widget/Toolbar.html) class to implement app bars as the most recent features are added to the support library's version of [Toolbar](https://developer.android.com/reference/android/support/v7/widget/Toolbar.html), and they are available on any device that can use the support library. 

I am working on it.

mejariamol:
fixes #128 
fixes #141 

**UPDATED:**

apk link - https://www.dropbox.com/s/2l2sgbu1oy8m2yg/app-debug.apk?dl=0

With this PR 

-  Added master/detail flow support for event list using fragments.

- Used support library's toolbar to implement action bar

- Removed `detach()` method from the presenters associated with the fragments since it gives **NPE** for **view** while switching between fragments. Also removed related tests.

- Modified some styling for `event_layout.xml`

screenshot-2017-05-31_17 42 53 359

screenshot_2017-05-31-16-06-39

screenshot-2017-05-30_14 49 07 593


mejariamol:
@iamareebjamal please review

iamareebjamal:
@mejariamol The item spacing in landscape tablet view is too large, can you reduce it please?

Also, in portrait mode, you can increase the left and right margin so that items appear more in center

mejariamol:
@iamareebjamal updated. :+1:

app/build.gradle
```diff
@@ -60,5 +57,10 @@ dependencies {
     compile 'de.hdodenhof:circleimageview:2.1.0'
     compile 'com.jakewharton.timber:timber:4.5.1'
     compile 'com.jakewharton:butterknife:8.5.1'
+    compile 'com.android.support:support-v4:25.3.1'
```
iamareebjamal: Please move the android support libraries to top and in 1 place. Similarly, test dependencies should be at top too.

Also, support-v4 is not needed as extra dependency. It's already included

app/src/main/java/org/fossasia/openevent/app/event/EventContainerFragment.java
```diff
@@ -0,0 +1,152 @@
...
+                    break;
+            }
+        }
+        fm.beginTransaction()
+            .replace(R.id.event_container, fg, tag)
+            .setTransition(FragmentTransaction.TRANSIT_FRAGMENT_FADE)
+            .commit();
+    }
+
+    // TODO: Rename method, update argument and hook method into UI event
```
iamareebjamal: Please fulfill the TODOs and remove them

app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesPresenter.java
```diff
@@ -37,11 +37,6 @@ public void attach() {
     }
 
     @Override
-    public void detach() {
```
iamareebjamal: Don't remove detach. We MUST detach presenter at some point, it is giving NPE just means it is detach at wrong lifecycle method

app/src/main/java/org/fossasia/openevent/app/event/detail/EventDetailsPresenter.java
```diff
@@ -36,11 +36,6 @@ public void attach() {
     }
 
     @Override
-    public void detach() {
```
iamareebjamal: Same here. If detach is removed, it opens new windows to numerous memory leaks. Please don't remove detach

app/src/main/java/org/fossasia/openevent/app/events/EventDetailActivity.java
```diff
@@ -0,0 +1,90 @@
...
+        ActionBar actionBar = getSupportActionBar();
+        if (actionBar != null) {
+            actionBar.setDisplayHomeAsUpEnabled(true);
+        }
+
+        initialEvent = getIntent().getParcelableExtra(EventListActivity.EVENT_KEY);
+
+        setTitle(initialEvent.getName());
+
+        // savedInstanceState is non-null when there is fragment state
```
iamareebjamal: This is self explanatory for any Android developer

Although comments are great, obvious ones just clutter the code.

Add comments which explain any custom behaviour implemented by you.

app/src/main/java/org/fossasia/openevent/app/events/EventDetailActivity.java
```diff
@@ -0,0 +1,90 @@
...
+                .add(R.id.event_detail_container, fg)
+                .commit();
+        }
+    }
+
+    @Override
+    public boolean onOptionsItemSelected(MenuItem item) {
+        int id = item.getItemId();
+        if (id == android.R.id.home) {
+            // This ID represents the Home or Up button. In the case of this
```
iamareebjamal: Same thing about this comment

app/src/main/java/org/fossasia/openevent/app/events/EventDetailActivity.java
```diff
@@ -0,0 +1,90 @@
...
+            // http://developer.android.com/design/patterns/navigation.html#up-vs-back
+            //
+            navigateUpTo(new Intent(this, EventListActivity.class));
+            return true;
+        }
+        return super.onOptionsItemSelected(item);
+    }
+
+    @Override
+    public void onFragmentInteraction(Uri uri) {
```
iamareebjamal: Why a new listener, I suppose that the current communication is working ?

app/src/main/java/org/fossasia/openevent/app/events/EventListActivity.java
```diff
@@ -29,55 +30,70 @@
...
-    ProgressBar progressBar;
+    @BindView(R.id.toolbar)
+    Toolbar toolbar;
 
-    public static List<Event> events = new ArrayList<>();
+    /**
+     * Whether or not the activity is in two-pane mode, i.e. running on a tablet
+     * device.
+     */
+    private boolean mTwoPane;
```
iamareebjamal: Don't use hungarian notation of variables. Only Google in the whole world does and for all wrong reasons

app/src/main/java/org/fossasia/openevent/app/events/EventListActivity.java
```diff
@@ -100,24 +116,24 @@ public boolean onOptionsItemSelected(MenuItem item) {
 
     @Override
     public void showProgressBar(boolean show) {
-        showView(progressBar, show);
+        // to be implemented again
```
iamareebjamal: Please complete the implementation with all tests passing

app/src/test/java/org/fossasia/openevent/app/presenter/AttendeePresenterTest.java
```diff
@@ -82,20 +82,6 @@ public void shouldLoadAttendeesAutomatically() {
     }
 
     @Test
-    public void shouldDetachViewOnStop() {
```
iamareebjamal: Never ever delete tests before completely being sure that it isn't needed. We need to always detach presenters at correct lifecycle method. Even for the MainActivity whose life span is complete lifespan of the application. Still need to detach it

iamareebjamal:
Also, please maintain the 1 PR per issue rule. This PR handles 2 issues.

app/src/main/java/org/fossasia/openevent/app/events/EventListActivity.java
iamareebjamal: Also, this is why need to add tests for views ASAP too. So that we fail the build if someone forgets to add appropriate implementation

app/src/main/java/org/fossasia/openevent/app/events/EventDetailActivity.java
iamareebjamal: Generally avoid adding unimplemented methods

iamareebjamal: It means they aren't needed currently and this pattern will create a lag in feature request/implementation

app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesPresenter.java
mejariamol: We were detaching presenter in `onDetach()` method of fragment which is last lifecycle method of fragment. And it is called every time while switching between fragments leads to NPE for view.
We can't detach the presenter in any of the fragment's lifecycle.

app/src/main/java/org/fossasia/openevent/app/events/EventDetailActivity.java
mejariamol: As this is open source I think such comments are always useful for new devs (in future) on board while getting start.
**P.S.** its auto generated and I think its nice :)

app/src/main/java/org/fossasia/openevent/app/events/EventDetailActivity.java
mejariamol: Yeah I will remove it. It is just that till now we didn't require any communication from fragment to the host.

mejariamol:
@iamareebjamal that was unintentional while reporting #141. I realized later the PR handling master/detail flow support will handle that too.

app/src/main/java/org/fossasia/openevent/app/events/EventDetailActivity.java
iamareebjamal: It's a standard Android knowledge, it isn't helpful to the project per se. If it were for beginners of Android, it would have been good. Our comments should ideally be not documenting(re) the standard Android behavior

app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesPresenter.java
iamareebjamal: We can detach, and then call on `onStart`, please try

mejariamol: the problem comes when fragment is switched while loading.
For example 
In `AttendeesPresenter` - 
```
eventDataRepository.getAttendees(eventId, forceReload)
            .subscribe(attendees -> {
                ...
                attendeesView.showAttendees(attendees);
                ...
            ...
            });
```
Once the presenter makes this call, while the subscriber is still waiting for the result, if fragment is switched that is presenter is detached then view becomes null and on the result emitted it throws NPE for `attendeesView`. 

Detaching presenter is only good in activity's `onDestroy`.

mejariamol: @iamareebjamal Well, let it be attached for now. Once we add **LeakCanary** we can raise it up as a separate issue.

iamareebjamal: Then please add a null check for subscriber as in all the presenters. It is needed for all subscribers as view can be null anyways even if you don't detach. Please do this and test

mejariamol:
@iamareebjamal please review now.

app/src/test/java/org/fossasia/openevent/app/presenter/EventsPresenterTest.java
```diff
@@ -135,7 +135,6 @@ public void shouldLoadOrganiserSuccessfully() {
 
         inOrder.verify(eventModel).getOrganiser(false);
         inOrder.verify(eventListView).showOrganiserName("John Wick");
-        inOrder.verify(eventListView).showOrganiserPanel(true);
```
iamareebjamal: Why removing this?

iamareebjamal:
Amazing work!

But please tell before modifying tests that why you need to do so

app/src/test/java/org/fossasia/openevent/app/presenter/EventsPresenterTest.java
iamareebjamal: I guess because it is now in Toolbar?

iamareebjamal: If so, please add a `Coordinator : ` prefix in the toolbar and make sure the string template is in strings.xml

mejariamol: @iamareebjamal done. :+1:

iamareebjamal:
@mejariamol Please always enable `Allow edits from maintainers`

I'm trying to push some tests and default behaviour, Please check PR in your branch

mejariamol:
@iamareebjamal yeah. done :+1:

PR #147, opened 2017-06-02
Add barcode detector frame on qr scan layout

### LINKED ISSUE #146: Add detector frame on QR scan layout
**UPDATED:**
Currently qr scan activity is just full screen activity. It gives no idea about what actually it is meant for.

Expected: 
- add detector frame on qr scan layout

I am working on it

mejariamol:
fixes #146 

screenshot_2017-06-02-09-17-50

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/app/qrscan/ScanQRActivity.java
+++ b/app/src/main/java/org/fossasia/openevent/app/qrscan/ScanQRActivity.java
@@ -4,13 +4,22 @@ import android.Manifest;
 import android.Manifest.permission;
 import android.content.Intent;
 import android.content.pm.PackageManager;
+import android.graphics.Point;
+import android.graphics.drawable.Drawable;
+import android.graphics.drawable.LayerDrawable;
 import android.os.Bundle;
 import android.support.annotation.NonNull;
+import android.support.design.widget.CoordinatorLayout;
 import android.support.v4.app.ActivityCompat;
 import android.support.v4.content.ContextCompat;
+import android.support.v4.content.res.ResourcesCompat;
 import android.support.v7.app.AppCompatActivity;
+import android.util.DisplayMetrics;
+import android.view.Display;
+import android.view.Gravity;
 import android.view.SurfaceHolder;
 import android.view.SurfaceView;
+import android.widget.FrameLayout;
 import android.widget.ProgressBar;
 import android.widget.TextView;
 import android.widget.Toast;
@@ -57,6 +66,18 @@ public class ScanQRActivity extends AppCompatActivity implements IScanQRView {
     @BindView(R.id.progressBar)
     ProgressBar progressBar;
 
+    @BindView(R.id.frame_corner_00)
+    FrameLayout frameCorner00;
+
+    @BindView(R.id.frame_corner_01)
+    FrameLayout frameCorner01;
+
+    @BindView(R.id.frame_corner_10)
+    FrameLayout frameCorner10;
+
+    @BindView(R.id.frame_corner_11)
+    FrameLayout frameCorner11;
+
     private CameraSource cameraSource;
     private BarcodeDetector barcodeDetector;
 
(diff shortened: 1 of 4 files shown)


mejariamol:
@iamareebjamal please review

app/src/main/java/org/fossasia/openevent/app/qrscan/ScanQRActivity.java
```diff
@@ -112,6 +135,66 @@ public void onRequestPermissionsResult(int requestCode, @NonNull String[] permis
 
     // Lifecycle methods end
 
+    private int convertDpToPixel(float dp) {
```
iamareebjamal: Move to View utils as static method

mejariamol:
Actually we need a **static frame** on the layout for always.

iamareebjamal:
The static frame like this implies that the camera will only scan codes inside this region. Which isn't true

mejariamol:
yes. We will be allowing full screen scan detection.

mejariamol:
>The static frame like this implies that the camera will only scan codes inside this region. Which isn't true

its not limiting detection area. But yes, capturing qr code in center is always advisable right?

mejariamol:
@iamareebjamal please review

iamareebjamal:
@mejariamol Please resolve conflicts

mejariamol:
@iamareebjamal done :+1:

iamareebjamal:
@mejariamol The current code is looking very hacky. Although I understand that we need this, do take a look at Vision APIs detectors and add that here instead because that was needed but was delegated for later. Once you add it, you can see for yourself if we need this. And if we do, you can create a custom view for camera frames which will be easy to make rather than setting 4 frames dynamically like this

mejariamol:
>do take a look at Vision APIs detectors and add that here instead

Yeah. I will open new issue for that. Then we will decide if we need this. :+1: 
>you can create a custom view for camera frames which will be easy to make rather than setting 4 frames dynamically like this

I didn't understand custom view. I am setting up framelayouts dynamically to adjust with all screen.

iamareebjamal:
@mejariamol True, but setting a customview would be better and more efficient than the current technique

app/src/main/res/layout/activity_scan_qr.xml
```diff
@@ -20,6 +20,42 @@
...
+        android:layout_gravity="center"
+        android:layout_marginBottom="160dp"
+        android:layout_marginLeft="160dp"
+        android:background="@drawable/bordered_rectangle"/>
+
+    <FrameLayout
+        android:id="@+id/frame_corner_10"
+        android:layout_width="50dp"
+        android:layout_height="50dp"
+        android:layout_gravity="center"
```
arp95: Nice work!
can you provide all these values in dimens.xml please.

arp95:
@mejariamol please squash the commits.

mejariamol:
@arp95 
>Its not working fluently

What's not working fluently?
btw we are not going for this. Instead google vision's barcode detector frame will be implemented. I am working on it.

arp95:
@mejariamol sure go ahead.

app/src/main/java/org/fossasia/openevent/app/qrscan/ScanQRActivity.java
```diff
@@ -5,14 +5,24 @@
...
+import android.graphics.drawable.LayerDrawable;
 import android.os.Bundle;
 import android.support.annotation.NonNull;
+import android.support.design.widget.CoordinatorLayout;
 import android.support.v4.app.ActivityCompat;
 import android.support.v4.content.ContextCompat;
+import android.support.v4.content.res.ResourcesCompat;
 import android.support.v7.app.AppCompatActivity;
+
+import android.util.DisplayMetrics;
```
rohanagarwal94: Please avoid unused imports as mentioned in the codacy review.

iamareebjamal: @rohanagarwal94 The PR is not active as mentioned above

mejariamol:
Will be handled in the new issue to add qr code detector frame provided by google vision api.

PR #153, opened 2017-06-03
feat: Add dependency injection using dagger

### LINKED ISSUE #145: Add Dagger2 for Dependency Injection
We have a great testing environment, but that also means that we have lots of class dependencies and have to configure them the same way in each object.

We should use Dagger 2 for DI

I'm working on it

iamareebjamal:
- Add dagger and configure it for different modules
-  Add start method in presenters instead of attach
- Modify tests accordingly
- Normalize model names

Fixes #145

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -62,7 +62,9 @@ dependencies {
     compile 'de.hdodenhof:circleimageview:2.1.0'
     compile 'com.jakewharton.timber:timber:4.5.1'
     compile 'com.jakewharton:butterknife:8.5.1'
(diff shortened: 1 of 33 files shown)


iamareebjamal:
@mejariamol Please review

Had to normalize several components so PR is big

iamareebjamal:
@mejariamol Please review as implementing dbflow on non dagger environment will be a bit problematic

mejariamol:
Yeah sure. I will review soon. I am getting familiar with it. :)

iamareebjamal:
@niranjan94 Please merge

PR #155, opened 2017-06-03
feat: Add data binding to layouts

### LINKED ISSUE #154: Use DataBinding to bind objects to views
Android DataBinding provides simple and efficient way to bind data to views and eliminate the need to add checks and formatting to many UI components.

It also provides ways of custom binding and 2 way binding and validation making it easy to write layouts with minimal expressions and decreasing the work done in presenters and views

iamareebjamal:
Fixes #154 

Add databinding to layouts containing data from models


iamareebjamal:
@mejariamol Please review

iamareebjamal:
There is no data to be bound there AFAIK?

app/src/main/java/org/fossasia/openevent/app/event/detail/EventDetailsPresenter.java
```diff
@@ -58,38 +60,43 @@ public void loadTickets(long eventId, boolean forceReload) {
...
 
-        eventDetailView.showTicketStats(totalTickets > 0 ? totalAttendees : 0, totalTickets);
+        event.totalTickets.set(totalTickets);
+        event.totalAttendees.set(totalAttendees);
+        event.checkedIn.set(checkedInAttendees);
 
         eventDetailView.showProgressBar(false);
     }
 
+    private void showEventInfo(Event event) {
```
mejariamol: change the method names to something like `setBindingEvent`, `setBindingEventInfo`. As actually we are setting up binding fields in these methods.

mejariamol: change the method name `setBindingEventInfo` or just `setBindingEvent`. As actually we are setting up binding fields in these methods.

app/src/main/java/org/fossasia/openevent/app/event/detail/EventDetailsPresenter.java
```diff
@@ -58,38 +60,43 @@ public void loadTickets(long eventId, boolean forceReload) {
...
+        if(eventDetailView == null)
```
mejariamol: I think now there is no need to check non-nullity of view. Please check if doing so is giving errors on null view. If not please remove this case.

app/src/main/java/org/fossasia/openevent/app/event/detail/EventDetailsPresenter.java
```diff
@@ -58,38 +60,43 @@ public void loadTickets(long eventId, boolean forceReload) {
...
+            return;
+
+        eventDetailView.showEvent(event);
```
mejariamol: Similarly `setBinding**` name convention here.

app/src/main/java/org/fossasia/openevent/app/event/detail/EventDetailsPresenter.java
```diff
@@ -106,8 +113,14 @@ public void loadAttendees(long eventId, boolean forceReload) {
     }
 
     private void processAttendeesAndDisplay(List<Attendee> attendees) {
```
mejariamol: `setBindingAttendees` would be enough.

app/src/main/java/org/fossasia/openevent/app/event/detail/EventDetailsPresenter.java
```diff
@@ -106,8 +113,14 @@ public void loadAttendees(long eventId, boolean forceReload) {
...
+        if(eventDetailView == null)
```
mejariamol: similarly check if we need to check this case. Else remove it.

app/src/main/java/org/fossasia/openevent/app/event/detail/EventDetailsFragment.java
```diff
@@ -110,39 +85,13 @@ public void onDetach() {
...
-    @Override
-    public void showEventName(String name) {
-        tvEventTitle.setText(name);
-    }
-
-    @Override
-    public void showDates(String start, String end) {
-        tvStartDate.setText(start);
-        tvEndDate.setText(end);
+        ViewUtils.showView(binding.progressBar, show);
```
mejariamol: Can't we use a boolean binding field to show the progressbar and setting it to `true` `false` accordingly.

app/src/main/java/org/fossasia/openevent/app/event/detail/EventDetailsPresenter.java
iamareebjamal: View shouldn't know what we are using as a technology. It should remain neutral. If we change databinding to something else, we shouldn't have to change contracts. Contracts always remain neutral

app/src/main/java/org/fossasia/openevent/app/event/detail/EventDetailsPresenter.java
iamareebjamal: There is a need. When the presenter is detached and a request takes time to return, it will cause memory leak. In short, the need of null check will always remain.

app/src/main/java/org/fossasia/openevent/app/event/detail/EventDetailsFragment.java
iamareebjamal: Yes. This is a proposed feature, where we will set a state to the model and change it to reflect in the binding, but currently we are changing the objects so a new object will override the state instead of toggling it. This is something we can work in future with better design on how to handle state changes

app/src/main/java/org/fossasia/openevent/app/event/detail/EventDetailsPresenter.java
iamareebjamal: View should be oblivious to the fact that we are using data binding. Contracts should remain general and not change if we change the technology. That's the point of contracts

iamareebjamal:
@niranjan94 Please merge

app/src/main/java/org/fossasia/openevent/app/events/EventsListAdapter.java
```diff
@@ -5,69 +5,41 @@
...
 
     @Override
     public void onBindViewHolder(final EventRecyclerViewHolder holder, int position) {
-
         final Event thisEvent = events.get(position);
-        holder.eventTitle.setText(thisEvent.getName());
-
-        String thumbnail = thisEvent.getThumbnail();
-
-        if(thumbnail != null && !TextUtils.isEmpty(thumbnail)) {
```
mejariamol: @iamareebjamal This checking is missed after the PR. Now app crashes in `EventsListActivity` on empty thumbnail field. Please make required changes.

app/src/main/java/org/fossasia/openevent/app/events/EventsListAdapter.java
```diff
@@ -5,69 +5,41 @@
...
-import android.view.View;
 import android.view.ViewGroup;
-import android.widget.ImageView;
-import android.widget.TextView;
-
-import com.squareup.picasso.Picasso;
 
 import org.fossasia.openevent.app.R;
 import org.fossasia.openevent.app.data.models.Event;
+import org.fossasia.openevent.app.databinding.EventLayoutBinding;
```
mejariamol: this is giving me error `Error:(13, 46) error: package org.fossasia.openevent.app.databinding does not exist`. Any hint?
@iamareebjamal

iamareebjamal:
@mejariamol if there's an error in any annotation processor. This error will come up. Please scroll down to see the actual error. This has nothing to do with data binding. Just with the fact that bindings aren't generated when there's an error in build

PR #159, opened 2017-06-07
fix: Add null check in image load binding adapter

### LINKED ISSUE #158: Crash on null url binding on image load
If thumbnail field is null, binding adapter does not check it and the app crashes

iamareebjamal:
Fixes #158 

Add null or empty check in binding adapter with Uri parse to ensure correct uri encoding


iamareebjamal:
@mejariamol @arp95 @mahikaw @Shailesh351 Please review

iamareebjamal:
@niranjan94 Please merge

PR #162, opened 2017-06-09
Event detail UI modification

### LINKED ISSUE #157: Redesign event detail layout
Current event detail layout can be modified into better one.

I am working on it.

**Proposed UI:**

Include both progressbar into a card view. Approximate view will be 
screenshot from 2017-06-07 15-16-34

@iamareebjamal please add `ui` label.

mejariamol:
fixes #157 

screenshot_2017-06-09-15-09-54

**P.S.** I have used hardcode dimension values because that will be taken up as a separate issue #163  with current hardcoded dimensions modification which is required.


mejariamol:

PR #164, opened 2017-06-09
modifies @dimens

### LINKED ISSUE #163: Modify @dimens
Current `dimens.xml` need to be modified.

I am working on it.

mejariamol:
fixes #163

### DIFF
--- a/app/src/main/res/values/dimens.xml
+++ b/app/src/main/res/values/dimens.xml
@@ -1,10 +1,18 @@
 <resources>
-    <!-- Default screen margins, per the Android Design guidelines. -->
-    <dimen name="activity_horizontal_margin">16dp</dimen>
-    <dimen name="activity_vertical_margin">16dp</dimen>
-    <dimen name="fab_margin">16dp</dimen>
-    <dimen name="app_bar_height">200dp</dimen>
-    <dimen name="item_width">200dp</dimen>
-    <dimen name="text_margin">16dp</dimen>
-    <dimen name="spacing_medium">16dp</dimen>
+
+    <dimen name="spacing_extra_small">4dp</dimen>
+    <dimen name="spacing_small">8dp</dimen>
+    <dimen name="spacing_medium">12dp</dimen>
+    <dimen name="spacing_normal">16dp</dimen>
+    <dimen name="spacing_large">32dp</dimen>
+    <dimen name="spacing_extra_large">64dp</dimen>
+
+    <dimen name="elevation">2dp</dimen>
+
+    <dimen name="text_size_small">12sp</dimen>
+    <dimen name="text_size_normal">16sp</dimen>
+    <dimen name="text_size_large">20sp</dimen>
+
+    <dimen name="image_small">50dp</dimen>
+
 </resources>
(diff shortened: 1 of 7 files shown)


mejariamol:
@iamareebjamal @arp95 @Shailesh351 @mahikaw @rohanagarwal94 Please review

app/src/main/res/values/dimens.xml
```diff
@@ -1,10 +1,25 @@
...
-    <!-- Default screen margins, per the Android Design guidelines. -->
-    <dimen name="activity_horizontal_margin">16dp</dimen>
-    <dimen name="activity_vertical_margin">16dp</dimen>
-    <dimen name="fab_margin">16dp</dimen>
-    <dimen name="app_bar_height">200dp</dimen>
-    <dimen name="item_width">200dp</dimen>
-    <dimen name="text_margin">16dp</dimen>
-    <dimen name="spacing_medium">16dp</dimen>
+
+    <dimen name="margin_extra_small">4dp</dimen>
```
iamareebjamal: Doesn't make sense to create different dimes for margins and padding. Create common `spacing_[medium,normal,small,...]` for use in both margin and padding

rohanagarwal94: @iamareebjamal I think its fair to use different dimensions for margins and padding. It just happens that they are same in the given case. However they might differ in the future.

arp95: i think you misunderstood @iamareebjamal .
what he meant is use general names for same padding or layout margin like small medium etc. i find that more appropriate.

mejariamol: @iamareebjamal @arp95 I have created separate padding and margin dimes intentionally. Both are very different things and will give different results for same values. 

for example - later we may change our padding scale. Like we may change `padding_normal` to `18dp` keeping `margin_normal` to `16dp`.

Hence keeping them separate is good idea and also its not gonna make any performance issues.;p

iamareebjamal: We can easily create a different spacing for that. I'm intentionally saying so they can be easily replaced and duplication is really bad. If we'll need to create new spacing, we'll do that.

The problem this way is that padding and margins are used interchangeably most of the time and they don't make a difference in most screens. If I quickly decide to change one dimen, it should be so across everything because material guidelines tell to keep constant spacings

rohanagarwal94: @iamareebjamal [padding] (https://github.com/google/iosched/blob/master/android/src/main/res/values/dimens.xml#L23) &
[margin](https://github.com/google/iosched/blob/master/android/src/main/res/values/dimens.xml#L218)
 rom google I/O app.

iamareebjamal: @rohanagarwal94 1 padding across all spacings :rofl: Come on! Furthermore, who said that google development standards are constant

Also, the margin you marked are default android activity margins ( Not supposed to touch them)

Besides you gave an excellect example of how to use spacing. Let's create spacing_[micro, normal, large]

rohanagarwal94: Glad I could be of help. Anyway spacing it is then :+1:

mejariamol:
@iamareebjamal can be merged now. :+1:

PR #186, opened 2017-06-13
fix: Load complete event on detail view load

iamareebjamal:
- Filter incomplete events in detail view
- Load from network if incomplete event found
- Batch insert mode through transaction

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/app/data/models/Event.java
+++ b/app/src/main/java/org/fossasia/openevent/app/data/models/Event.java
@@ -99,7 +99,7 @@ public class Event extends BaseModel implements Parcelable {
     private String organizerName;
 
     // Tells if the event saved is complete ( with tickets )
-    private boolean isComplete = true;
+    private boolean isComplete;
 
     // For Data Binding
     public final ObservableField<String> startDate = new ObservableField<>();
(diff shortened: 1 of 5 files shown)


iamareebjamal:
@arp95 @mejariamol @Shailesh351 @rohanagarwal94 Please review

app/src/main/java/org/fossasia/openevent/app/data/models/Event.java
```diff
@@ -99,7 +99,7 @@
     private String organizerName;
 
     // Tells if the event saved is complete ( with tickets )
-    private boolean isComplete = true;
+    private boolean isComplete;
```
mejariamol: Prefer default value to `false` and set it to true at appropriate place.

iamareebjamal: That is what I have done here

mejariamol: You are setting it to false while downloading events for the first time. I am saying to let the default value be `false` in Event.class` itself. And it can be set to true at appropriate place.

iamareebjamal: I can't see where I have done it. Please point out

mejariamol: Sorry. I might have read it in old code. All gud :+1:

iamareebjamal:
@niranjan94 Please merge

PR #192, opened 2017-06-15
Retains fragment instances for same event

### LINKED ISSUE #191: Event containing fragments are not retained
fragment instances are not retained while switching for same event.

Expected: fragments must be retained for same event

mejariamol:
fixes #191


mejariamol:
@iamareebjamal please review

iamareebjamal:
@mejariamol Please test now and see if the log messages for loading attendees display on changing of tabs. I'm currently AFK

mejariamol:
@iamareebjamal yes. I debugged. Now fragment instances are retained. But the attendees will be loaded everytime (from network initially and from disk thereafter) since presenter's `start()` call is called in `onActivityCreated` which is no issue.

iamareebjamal:
But they shouldn't load on tab switches.

Well, I'll think about it later and fix it

mejariamol:
For that to happen presenter's data loading calls should be separated from view's interactions, And these shall be called in fragment's `onCreate`.

PR #195, opened 2017-06-15
Fixes progressbar on event detail page

### LINKED ISSUE #194: Progressbar doesn't show up on event detail page
Progress bar doesn't show up on event detail page while data is still loading

Expected: Progress bar must be shown until page is loaded with complete data.

mejariamol:
fixes #194

There are two async calls on event detail page. hence progressbar was getting hidden by one of them while other is in progress.

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/app/event/detail/EventDetailPresenter.java
+++ b/app/src/main/java/org/fossasia/openevent/app/event/detail/EventDetailPresenter.java
@@ -25,6 +25,14 @@ public class EventDetailPresenter implements IEventDetailPresenter {
     private long totalAttendees;
     private long checkedInAttendees;
 
+    /**
+     * progress is parameter to check if complete data is loaded.
+     * we have two async processes loadAttendees and loadTickets.
+     * on completion of each one progress will be incremented
+     * hence progressbar will be hidden when progress is greater than equal to 2
+     */
+    private int progress = 0;
+
     @Inject
     public EventDetailPresenter(IEventRepository eventRepository) {
         this.eventRepository = eventRepository;
@@ -38,6 +46,7 @@ public class EventDetailPresenter implements IEventDetailPresenter {
 
     @Override
     public void start() {
+        progress = 0;
         showEventInfo(initialEvent);
 
         loadAttendees(initialEvent.getId(), false);
@@ -63,7 +72,7 @@ public class EventDetailPresenter implements IEventDetailPresenter {
                     if(eventDetailView == null)
                         return;
                     eventDetailView.showEventLoadError(throwable.getMessage());
-                    eventDetailView.showProgressBar(false);
+                    hideProgressbar();
                 });
     }
 
@@ -87,7 +96,7 @@ public class EventDetailPresenter implements IEventDetailPresenter {
         event.totalAttendees.set(totalAttendees);
         event.checkedIn.set(checkedInAttendees);
 
-        eventDetailView.showProgressBar(false);
+        hideProgressbar();
     }
 
     private void showEventInfo(Event event) {
@@ -109,6 +118,8 @@ public class EventDetailPresenter implements IEventDetailPresenter {
         if(eventDetailView == null)
             return;
 
+        eventDetailView.showProgressBar(true);
+
         eventRepository
             .getAttendees(eventId, forceReload)
             .toList()
@@ -117,6 +128,8 @@ public class EventDetailPresenter implements IEventDetailPresenter {
                     if(eventDetailView == null)
                         return;
                     eventDetailView.showEventLoadError(throwable.getMessage());
+
+                    hideProgressbar();
                 });
     }
 
@@ -143,7 +156,26 @@ public class EventDetailPresenter implements IEventDetailPresenter {
 
                 if(event != null)
                     event.checkedIn.set(checkedInAttendees);
+
+                hideProgressbar();
             });
+    }
+
+    /**
+     * checks if complete data is loaded
+     * and hides progressbar accordingly
+     */
+    private boolean hideProgressbar() {
+        progress ++;
+        if (progress >= 2) {
+            eventDetailView.showProgressBar(false);
+            return true;
+        }
+        return false;
+    }
+
+    public int getProgress() {
+        return progress;
     }
 
     public IEventDetailView getView() {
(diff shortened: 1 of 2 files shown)


mejariamol:

app/src/main/java/org/fossasia/openevent/app/event/detail/EventDetailPresenter.java
```diff
@@ -143,6 +162,10 @@ private void processAttendeesAndDisplay(List<Attendee> attendees) {
 
                 if(event != null)
                     event.checkedIn.set(checkedInAttendees);
+
+                progress ++;
```
iamareebjamal: This code block is repeating, please create function for these 3 lines

app/src/test/java/org/fossasia/openevent/app/presenter/EventDetailPresenterTest.java
```diff
@@ -130,7 +130,6 @@ public void shouldShowEventError() {
         inOrder.verify(eventDetailView).showProgressBar(true);
         inOrder.verify(eventRepository).getEvent(id, false);
         inOrder.verify(eventDetailView).showEventLoadError(error);
-        inOrder.verify(eventDetailView).showProgressBar(false);
```
iamareebjamal: You just deleted a test for hiding the progress bar. Now anyone can remove the hiding code and be done with it because the tests will not fail. Please add a test which ensures that if someone forgets to hide the progress bar, it will fail and test it by forgetting to hide the progress bar and confirming that it fails

mejariamol: Test should test app's behavior not what developer has coded. I will add test to check if progressbar gets hidden when complete data gets loaded. :+1:

iamareebjamal: That is exactly what I was implying

mejariamol: @iamareebjamal done :+1: please review

app/src/test/java/org/fossasia/openevent/app/presenter/EventDetailPresenterTest.java
```diff
@@ -192,4 +190,23 @@ public void shouldNotAccessView() {
...
+
+        when(eventRepository.getEvent(id, false))
+            .thenReturn(Observable.just(event));
+
+        InOrder inOrder = Mockito.inOrder(eventRepository, eventDetailView);
+
+        eventDetailPresenter.loadAttendees(id, false);
+        eventDetailPresenter.loadTickets(id, false);
+
+        while(eventDetailPresenter.getProgress() < 2) {
```
iamareebjamal: This is not the correct way to write the test. Load both the attendees and you should not know that the progress should be less than 2 and I have to wait it it becomes so. Besides, all the work is synchronous in threads, so it should work after removing this

app/src/test/java/org/fossasia/openevent/app/presenter/EventDetailPresenterTest.java
mejariamol: @iamareebjamal please check now.

app/src/test/java/org/fossasia/openevent/app/presenter/EventDetailPresenterTest.java
```diff
@@ -192,4 +192,19 @@ public void shouldNotAccessView() {
...
+
+    @Test
+    public void shouldHideProgressbarCorrectly() {
+        when(eventRepository.getAttendees(id, false))
+            .thenReturn(Observable.fromIterable(attendees));
+
+        when(eventRepository.getEvent(id, false))
+            .thenReturn(Observable.just(event));
+
+        InOrder inOrder = Mockito.inOrder(eventRepository, eventDetailView);
```
iamareebjamal: There's no transaction with eventRepository, so please remove it from in order, and move the progress bar showing hiding to single test and remove the show code from above two tests

app/src/test/java/org/fossasia/openevent/app/presenter/EventDetailPresenterTest.java
```diff
@@ -192,4 +191,18 @@ public void shouldNotAccessView() {
...
+        when(eventRepository.getAttendees(id, false))
+            .thenReturn(Observable.fromIterable(attendees));
+
+        when(eventRepository.getEvent(id, false))
+            .thenReturn(Observable.just(event));
+
+        eventDetailPresenter.start();
+
+        assertTrue(eventDetailPresenter.getProgress() >= 2);
+        verify(eventDetailView).showProgressBar(false);
```
iamareebjamal: Move the progress bar showing hiding to single test and remove the show code from above two tests

mejariamol: I have did the same. I guess. :)

iamareebjamal: This is only verifying that the progress bar was hidden. And the shown test is in two separate tests. I  was talking about moving the show and hide test in single one and remove the show test from those two

mejariamol: please check now. @iamareebjamal

iamareebjamal:
@arp95 @mahikaw @Shailesh351 @rohanagarwal94 @shreyasnbhat Please review

arp95:
@iamareebjamal @mejariamol
cant we something similar to open event android app? i mean showing percentage of downloads as well?

iamareebjamal:
@arp95 This is for organizers. Don't think they'll mean anything to clients

app/src/test/java/org/fossasia/openevent/app/presenter/EventDetailPresenterTest.java
```diff
@@ -192,4 +190,22 @@ public void shouldNotAccessView() {
...
+        when(eventRepository.getEvent(id, false))
+            .thenReturn(Observable.just(event));
+
+        InOrder inOrder = Mockito.inOrder(eventDetailView);
+
+        eventDetailPresenter.start();
+
+        assertTrue(eventDetailPresenter.getProgress() >= 2);
+
+        inOrder.verify(eventDetailView, atLeast(2)).showProgressBar(true);
```
iamareebjamal: At least once please. 2 is a side effect of our implementation. If it is improved, we'll only be showing it once.

mejariamol: True :+1:

PR #204, opened 2017-06-19
feat: Change API according FastAdapter and DBFlow

### LINKED ISSUE #203: Make model and object flow compatible with current data flow for Database and FastAdapter
- Models which are used in FastAdapter are extending from BaseModel, however, our API works well without extending from it. They should be generified to reduce boilerplate
- Now that we are working through Database, parceling and unparceling of data is not required and should be propagated through DB API instead
- The methods created in FastAdapter use raw types, causing warnings and possible errors in future

iamareebjamal:
- Change DatabaseRepository API to be more general and less BaseModel specific
- Remove constraint on BaseModel for models used in FastAdapter and rather extend from AbstractItem
- Remove Parcelable models and propagate data through keys
- Remove raw types and add correct generics in FastAdapter classes

Fixes #203

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/app/common/BaseFragment.java
+++ b/app/src/main/java/org/fossasia/openevent/app/common/BaseFragment.java
@@ -1,12 +1,32 @@
 package org.fossasia.openevent.app.common;
 
+import android.app.Activity;
 import android.support.v4.app.Fragment;
+import android.support.v7.app.ActionBar;
+import android.support.v7.app.AppCompatActivity;
 
 import com.squareup.leakcanary.RefWatcher;
 
 import org.fossasia.openevent.app.OrgaApplication;
 
+import timber.log.Timber;
+
 public abstract class BaseFragment extends Fragment {
+
+    protected void setTitle(String title) {
+        Activity activity = getActivity();
+
+        if (activity != null && activity instanceof AppCompatActivity) {
+            ActionBar actionBar = ((AppCompatActivity) activity).getSupportActionBar();
+            if (actionBar != null)
+                actionBar.setTitle(title);
+            else
+                Timber.e("No ActionBar found in Activity %s for Fragment %s", activity, this);
+        } else {
+            Timber.e("Fragment %s is not attached to any Activity", this);
+        }
+
+    }
 
     @Override public void onDestroy() {
         super.onDestroy();
--- a/app/src/main/java/org/fossasia/openevent/app/data/models/Attendee.java
+++ b/app/src/main/java/org/fossasia/openevent/app/data/models/Attendee.java
@@ -1,41 +1,39 @@
 package org.fossasia.openevent.app.data.models;
 
-import android.content.Context;
-import android.os.Parcel;
-import android.os.Parcelable;
+import android.databinding.DataBindingUtil;
 import android.support.annotation.NonNull;
-import android.view.LayoutInflater;
 import android.view.View;
-import android.view.ViewGroup;
 
 import com.google.gson.annotations.SerializedName;
-import com.mikepenz.fastadapter.IItem;
+import com.mikepenz.fastadapter.items.AbstractItem;
+import com.raizlabs.android.dbflow.annotation.Column;
 import com.raizlabs.android.dbflow.annotation.ForeignKey;
 import com.raizlabs.android.dbflow.annotation.ForeignKeyAction;
 import com.raizlabs.android.dbflow.annotation.PrimaryKey;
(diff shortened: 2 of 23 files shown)


app/src/main/java/org/fossasia/openevent/app/common/BaseFragment.java
```diff
@@ -1,13 +1,33 @@
...
 
 import com.squareup.leakcanary.RefWatcher;
 
 import org.fossasia.openevent.app.OrgaApplication;
 
+import timber.log.Timber;
+
 public abstract class BaseFragment extends Fragment {
 
+    protected void setTitle(String title) {
```
mejariamol: Don't do this as it will be problem for two pane mode. (As we don't want to change the action bar title in two pane mode)
For one pane mode, this has been already taken care at single place in `EventDetailActivity`

app/src/main/java/org/fossasia/openevent/app/data/models/Attendee.java
```diff
@@ -253,83 +172,19 @@ public int getLayoutRes() {
...
 
     @Override
     public void bindView(AttendeeViewHolder holder, List<Object> list) {
+        super.bindView(holder, list);
         holder.bindAttendee(this);
     }
 
     @Override
     public void unbindView(AttendeeViewHolder holder) {
+        super.unbindView(holder);
```
mejariamol: this got nothing

app/src/main/java/org/fossasia/openevent/app/event/detail/EventDetailFragment.java
```diff
@@ -95,6 +96,7 @@ public void showProgressBar(boolean show) {
 
     @Override
     public void showEvent(Event event) {
+        setTitle(event.getName());
```
mejariamol: hence please remove this.

app/src/main/java/org/fossasia/openevent/app/common/BaseFragment.java
iamareebjamal: That was removed since we are now dealing with IDs only and not objects

app/src/main/java/org/fossasia/openevent/app/data/models/Attendee.java
iamareebjamal: What do you mean?

mejariamol: it is empty in super.

app/src/main/java/org/fossasia/openevent/app/common/BaseFragment.java
mejariamol: Or you can just send event name to EventDetailActivity from EventListActivity

iamareebjamal: Good idea

app/src/main/java/org/fossasia/openevent/app/data/models/Attendee.java
iamareebjamal: It requires you to override and call this

app/src/main/java/org/fossasia/openevent/app/common/BaseFragment.java
iamareebjamal: I'm leaving the method for future use

mejariamol:
please check the test

iamareebjamal:
@mejariamol 

```Java
public void shouldToggleSortedAttendeesSuccessfully() {
        Attendee targetAttendee = attendees.get(3);

        when(eventModel.toggleAttendeeCheckStatus(id, targetAttendee.getId()))
            .thenReturn(Observable.just(targetAttendee));

        // Set the shuffled list to search into
        List<Attendee> shuffled = new ArrayList<>(attendees);
        Collections.shuffle(shuffled);

        attendeesPresenter.setAttendeeList(shuffled);
        attendeesPresenter.toggleAttendeeCheckStatus(targetAttendee);

        Mockito.verify(attendeesView).updateAttendee(shuffled.indexOf(targetAttendee), targetAttendee);
    }
```

This is a flaky test. We should not randomize or shuffle in tests. Tests are statically configured. Besides there is no need to shuffle it, because the method it used to determine the index is not static, and is already unit tested against all such conditions. 

And it is failing because the shuffled List is returning the wrong index (shockingly), not because our methods are wrong

iamareebjamal:
@arp95 @rohanagarwal94 @Shailesh351 Please review

iamareebjamal:
@arp95 @rohanagarwal94 @Shailesh351 Please review

PR #212, opened 2017-06-21
feat: Load complete event on login of user

### LINKED ISSUE #205: App crash if attendees are saved before complete event
As attendee rely on the ticket to be foreign key, Foreign Key Constraint fails when attendees are saved before tickets( included in complete event )

We have to design a synchronisation mechanism to ensure this doesn't happen

iamareebjamal:
Fixes #205 

- Load the complete event with the partial event
- Add corresponding test

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/app/data/EventRepository.java
+++ b/app/src/main/java/org/fossasia/openevent/app/data/EventRepository.java
@@ -110,9 +110,21 @@ public class EventRepository implements IEventRepository {
 
         Observable<Event> networkObservable = Observable.defer(() ->
             eventService.getEvents(authorization)
-                .doOnNext(events -> databaseRepository.saveList(Event.class, events)
+                .doOnNext(events ->
+                    databaseRepository.saveList(Event.class, events)
                     .subscribe())
-                .flatMapIterable(events -> events));
+                .flatMapIterable(events -> events))
+                .doOnEach(eventNotification -> {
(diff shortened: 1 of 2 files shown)


iamareebjamal:
@mejariamol @rohanagarwal94 @arp95 @Shailesh351 Please review

app/src/main/java/org/fossasia/openevent/app/data/EventRepository.java
```diff
@@ -110,9 +110,21 @@ public EventRepository(IUtilModel utilModel, IDatabaseRepository databaseReposit
...
+                    databaseRepository.saveList(Event.class, events)
                     .subscribe())
-                .flatMapIterable(events -> events));
+                .flatMapIterable(events -> events))
+                .doOnEach(eventNotification -> {
+                    // Download all complete events in one go
+                    if (!eventNotification.isOnNext())
+                        return;
+                    Event event = eventNotification.getValue();
+                    getEvent(event.getId(), false)
```
mejariamol: Please add check here if an event is incomplete.

iamareebjamal: getEvent always returns complete event. Checking completeness is redundant

mejariamol: I was saying before calling `getEvent`. But I think its not required now. What do you say?

PR #215, opened 2017-06-21
fix: Clear database and token cache on different user login

### LINKED ISSUE #210: Clear all the database on new user login
Currently we are keeping database excluding `User` and `UserDetail`tables on log out. Which is problem for new login as it still shows old data for new user.

Correct approach should be keeping all the database on logout. On login if data exist check for the user id. If user id of the new user doesn't matches with existing id then clear all the database otherwise keep the same.

@iamareebjamal you may want to take up this.

iamareebjamal:
Fixes #210 

- Clear database on new user login
- Invalidate token cache from appropriate locations
- Add test

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/app/data/UtilModel.java
+++ b/app/src/main/java/org/fossasia/openevent/app/data/UtilModel.java
@@ -4,9 +4,14 @@ import android.content.Context;
 import android.content.SharedPreferences;
 import android.support.annotation.StringRes;
 
+import org.fossasia.openevent.app.OrgaApplication;
 import org.fossasia.openevent.app.data.contract.IUtilModel;
+import org.fossasia.openevent.app.data.db.configuration.OrgaDatabase;
 import org.fossasia.openevent.app.utils.Constants;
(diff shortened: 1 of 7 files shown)


iamareebjamal:
@mejariamol @arp95 @Shailesh351 @rohanagarwal94 Please review

mariobehling:
Please resolve conflicts.

app/src/main/java/org/fossasia/openevent/app/data/UtilModel.java
```diff
@@ -44,18 +49,33 @@ public void saveString(String key, String value) {
...
         saveString(Constants.SHARED_PREFS_TOKEN, token);
     }
 
     @Override
     public boolean isConnected() {
         return NetworkUtils.isNetworkConnected(context);
     }
+
+    @Override
+    public Completable deleteDatabase() {
```
mejariamol: Can we test this if it is actually deleting database?

iamareebjamal: Unit test or manual testing. For manual testing, you can check stetho

Issue #196, opened 2017-06-15
Separate initial data load part from ui interation in presenters


mejariamol:
Currently in presenters of event containing fragments, initial data loading part is in conjunction with a part interacting with the view. Hence data loading cannot be initialized in `onCreate` fragment call. Hence data load is called every time when switching between fragments.

If we manage to separate initial data loading as separate functions and maintain a flag to check progress and show result in view accordingly, it can be solved.

@iamareebjamal what do you think?

iamareebjamal:
True, but let's focus on #161 first. I have been creating mock models and API. No one has reported anything related to it. It is a core feature. Please let's focus on it. Follow up in https://github.com/fossasia/open-event-android/issues/1680

mejariamol:
saving and providing presenters will be done via this issue right? @iamareebjamal

iamareebjamal:
Yes

iamareebjamal:
It will be done by loaders

mejariamol:
taking this up

iamareebjamal:
Try to handle it in BaseFragment so you don't have to do it everywhere

Issue #148, opened 2017-06-02
Make event container page starter page after log in


mejariamol:
Currently after log in event list page is a starter page but for event organizer, starter page should be event container page for a selected event. And from navigation drawer organizer can go to the event list page and select the required event. 

@niranjan94 @iamareebjamal what is your view ?

iamareebjamal:
How would you choose which event to show as first? I think current UX is correct

mejariamol:
That is for now since we are not maintaining local database yet. Once we do that we will be having current selected event. 
For now if no event is selected we can redirect to the event list page.

iamareebjamal:
I guess that's too much complication for a basic feature. It's also not just a event page redirection problem, it is essentially a state saving problem. So let's tackle it in that manner only.

Let's say an organizer exits the application and returns back to the app and sees the last event he opened. I think he is most likely to get confused as he clearly pressed back, came to home screen and pressed back again to exit, so when the flow of app is to list->select->detail, why showing detail immediately?

mejariamol:
>why showing detail immediately?

Because its **NOT** like at a time organizer will be handling more than one event. At a time there will be mostly one event and it will be tedious for organizer to always check the event from list on starting.
The focus of the app should be at event detail not on the list.

iamareebjamal:
We can do this -> If there is one event. Show it directly. Sounds good?

mejariamol:
You are not getting my point. Please see the eventbrite organizer app on the playstore.

mejariamol:
@iamareebjamal did you check [eventbrite app](https://play.google.com/store/apps/details?id=com.eventbrite.organizer)? It is cleverly designed. I think we can take ideas from the design.

iamareebjamal:
OK I'm getting it, but let the development flow be organic and dynamic. Let us create a persistence layer first and then discuss this?

mejariamol:
@iamareebjamal please label it as `later`

mejariamol:
As now we have set up data persistence layer, I think we can take up this.

Expected:
After log in, Main page should be event detail page with navigation drawer. User can navigate to the sections of event via navigation drawer. Also event list can be approached from navigation drawer to select any event. Like in eventbrite app - 

![screenshot_2017-06-19-22-34-01](https://user-images.githubusercontent.com/13533840/27296853-410fc994-5540-11e7-9dd4-d7750a39a2fb.png)

[...]

PR #216, opened 2017-06-21
feat: Add swipe to refresh in all views

### LINKED ISSUE #211: Add swipe to refresh to refresh stale data
Add a swipe to refresh option on events, event detail, and attendees page

iamareebjamal:
Fixes #211 

- Add pull to refresh in all views
- Modify tests accordingly

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/app/event/EventContainerFragment.java
+++ b/app/src/main/java/org/fossasia/openevent/app/event/EventContainerFragment.java
@@ -74,6 +74,9 @@ public class EventContainerFragment extends BaseFragment {
         // Inflate the layout for this fragment
         View view = inflater.inflate(R.layout.fragment_event_container, container, false);
         ButterKnife.bind(this, view);
+
+        setRetainInstance(true);
+
         return view;
     }
 
--- a/app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesFragment.java
+++ b/app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesFragment.java
@@ -5,7 +5,9 @@ import android.content.Intent;
 import android.databinding.DataBindingUtil;
 import android.os.Bundle;
 import android.support.annotation.Nullable;
+import android.support.design.widget.Snackbar;
 import android.support.v4.app.Fragment;
+import android.support.v4.widget.SwipeRefreshLayout;
 import android.support.v7.app.AlertDialog;
 import android.support.v7.widget.DefaultItemAnimator;
 import android.support.v7.widget.DividerItemDecoration;
@@ -25,6 +27,7 @@ import com.timehop.stickyheadersrecyclerview.StickyRecyclerHeadersDecoration;
 import org.fossasia.openevent.app.OrgaApplication;
 import org.fossasia.openevent.app.R;
 import org.fossasia.openevent.app.common.BaseFragment;
+import org.fossasia.openevent.app.data.contract.IUtilModel;
 import org.fossasia.openevent.app.data.models.Attendee;
 import org.fossasia.openevent.app.databinding.FragmentAttendeesBinding;
 import org.fossasia.openevent.app.event.attendees.contract.IAttendeesPresenter;
@@ -57,9 +60,15 @@ public class AttendeesFragment extends BaseFragment implements IAttendeesView {
     public static final int REQ_CODE = 123;
 
     @Inject
+    IUtilModel utilModel;
+
+    @Inject
     IAttendeesPresenter attendeesPresenter;
 
+    private StickyHeaderAdapter stickyHeaderAdapter;
+    private RecyclerView.AdapterDataObserver adapterDataObserver;
     private FragmentAttendeesBinding binding;
+    private SwipeRefreshLayout refreshLayout;
 
     public AttendeesFragment() {
         // Required empty public constructor
@@ -79,6 +88,8 @@ public class AttendeesFragment extends BaseFragment implements IAttendeesView {
         fragment.setArguments(args);
         return fragment;
     }
+
+    // Lifecycle methods
 
     @Override
     public void onCreate(Bundle savedInstanceState) {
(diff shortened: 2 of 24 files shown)


iamareebjamal:
@mejariamol @Shailesh351 @arp95 @rohanagarwal94 Please review

app/src/main/java/org/fossasia/openevent/app/event/EventContainerFragment.java
```diff
@@ -74,6 +74,9 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container,
         // Inflate the layout for this fragment
         View view = inflater.inflate(R.layout.fragment_event_container, container, false);
         ButterKnife.bind(this, view);
+
+        setRetainInstance(true);
```
mejariamol: what is actual purpose of this?
If it is required, it should be called in `onCreate` instead.

iamareebjamal: It can be called anywhere and creating another lifecycle method becomes unmanageable, that's why I declared in onCreateView

app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesFragment.java
```diff
@@ -164,6 +155,43 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
...
+                decoration.invalidateHeaders();
+            }
+        };
+        stickyHeaderAdapter.registerAdapterDataObserver(adapterDataObserver);
+    }
+
+    private void setupRefreshListener() {
+        refreshLayout = binding.swipeContainer;
+        refreshLayout.setColorSchemeColors(utilModel.getResourceColor(R.color.colorAccent));
+        refreshLayout.setOnRefreshListener(() -> attendeesPresenter.loadAttendees(true));
```
mejariamol: hide appearing progress bar here as refresh layout already has its own.

app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesFragment.java
```diff
@@ -103,6 +114,7 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container,
         binding = DataBindingUtil.inflate(inflater, R.layout.fragment_attendees, container, false);
 
         binding.setAttendees(attendeesPresenter.getAttendees());
+        setRetainInstance(true);
```
mejariamol: Please remove this. It is causing issue while screen rotation
On screen rotation all loaded data in the layout vanishes.

app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesFragment.java
iamareebjamal: We have defined the progress to show whenever the view is updating. Be it refresh or loading. Hiding progressbar here would not be a great idea as it'll hinder the current workflow. Besides, maybe in future, we'll disable the UI while updating, so a progressbar will be a cue to that. The refresh layout just shows that it's a refresh action

app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesFragment.java
iamareebjamal: Not working if I remove this too. I checked. This way at least it is not showing 404 not found. And we need to retain fragments. Please open an issue for this. This bug is even in master right now

mejariamol: I removed this just from `AttendeesFragment`, and its working for me.

mejariamol: Please try removing it from `AttendeesFragment` and `EventDetailFragment`. Keeping in `EventContainerFragment`

iamareebjamal: Ok. Doing it, but create an issue for retaining fragments, we don't want reloading to happen on rotation

mejariamol: Its not reloading on rotation. with this 
> removing it from AttendeesFragment and EventDetailFragment. but Keeping in EventContainerFragment

iamareebjamal: device-2017-06-22-003731

No, it's not

mejariamol: May be because we are keeping attendees list with presenter it stays. hence works for attendees fragment

iamareebjamal: Got the error, fixing

mejariamol: Created issue for it #218

iamareebjamal:
@mejariamol Please check now. Data is still reloading from database on rotation. I'll create an issue

iamareebjamal:
@mejariamol Please approve

app/src/main/java/org/fossasia/openevent/app/event/detail/EventDetailPresenter.java
```diff
@@ -46,12 +48,27 @@ public void attach(IEventDetailView eventDetailView, long initialEventId) {
         this.initialEventId = initialEventId;
     }
 
+    private void loadAll(boolean refresh) {
```
mejariamol: this can be shifted in interface method `start()` making it `start(boolean forceStart)` removing `refresh()`

iamareebjamal: Currently, we have this complex way to load both attendees and event which will change once you complete #148, so we won't even have refresh. Besides, the idea of start is to keep it consistent through all presenters, maybe I should make all presenters extend an interface with start and detach in them. Doing this so that if we create a general class to maintain lifecycle or use Android Architecture Components, we don't have to implement it in each class, so start, attach and detach should have same parameters throughout the framework

mejariamol: >we have this complex way to load both attendees and event which will change once you complete #148 

Are we going for that structure? I thought we decided to keep present structure.

iamareebjamal: That is ideal in future when we have more than just attendees to show, but the point is that start, detach and attach should be consistent across presenters

mejariamol: Thats fine. But I think `start()` method of presenter should carry a parameter which will notify presenter if it needs to load data from disk. Currently every time start is called presenter loads data again.

Also this will fix #196 
@iamareebjamal what do you say

iamareebjamal: I think presenter should not even reset when orientation change happens. Many MVP libraries have implemented this. I think we should focus on presenter view implementation itself. Including a param will be a workaround to this problem, not a solution, not to mention that then in future, we'll have to implement it in every presenter. This will break DRY (Don't repeat yourself). That same if else everywhere. I'm studying this https://github.com/quangctkm9207/mvp-android-arch-component to apply architecture components in MVP, so it'll solve both our problems

iamareebjamal:
@niranjan94 Please merge

mariobehling:
@iamareebjamal Please try to get two reviews from co-developers where-ever possible.

iamareebjamal:
@mariobehling I tried to ping them twice

PR #219, opened 2017-06-22
fix: Retain fragment on screen rotation [WIP]

### LINKED ISSUE #218: Handle screen rotation
Currently on screen rotation event detail fragment looses its data.

Expected: data must be retained on screen rotation

**The problem is to retain more than one fragment on screen rotation.**
- `setRetainInstance(true)`
is one option to retain fragment instance on screen rotation that is on `onDetach()` call.
But using this only one object can be retained in the activity. We have 3 fragments currently. 
    - **EventContainerFragment** (direct fragment to the host activity - EventDetailActivity)
    - **AttendeesFragment** (child fragment of EventContainerFragment)
    - **EventDetailFragment** (child fragment of EventContainerFragment)

So we have to retain all these child fragments on screen rotation to avoid repetitive loading from the disk.

mejariamol:
- modifies presenter start method to check if data loaded

fixes #218

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesFragment.java
+++ b/app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesFragment.java
@@ -1,6 +1,5 @@
 package org.fossasia.openevent.app.event.attendees;
 
-import android.content.Context;
 import android.content.Intent;
 import android.databinding.DataBindingUtil;
 import android.os.Bundle;
@@ -48,8 +47,6 @@ import timber.log.Timber;
  */
 public class AttendeesFragment extends BaseFragment implements IAttendeesView {
 
-    private Context context;
-
     private long eventId;
 
     private ItemAdapter<Attendee> itemAdapter;
@@ -74,26 +71,24 @@ public class AttendeesFragment extends BaseFragment implements IAttendeesView {
      */
     public static AttendeesFragment newInstance(long eventId) {
         AttendeesFragment fragment = new AttendeesFragment();
-        Bundle args = new Bundle();
-        args.putLong(EventListActivity.EVENT_KEY, eventId);
-        fragment.setArguments(args);
+        fragment.setEventId(eventId);
         return fragment;
+    }
+
+    public void setEventId(long eventId) {
+        this.eventId = eventId;
     }
 
     @Override
     public void onCreate(Bundle savedInstanceState) {
-        context = getActivity();
 
         OrgaApplication
-            .getAppComponent(context)
+            .getAppComponent(getActivity())
             .inject(this);
 
         super.onCreate(savedInstanceState);
 
-        if (getArguments() != null) {
-            eventId = getArguments().getLong(EventListActivity.EVENT_KEY);
-            attendeesPresenter.attach(eventId, this);
-        }
+        setRetainInstance(true);
     }
 
     @Override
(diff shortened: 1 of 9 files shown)


mejariamol:
@iamareebjamal Working well with event detail fragment. For attendees fragment event list activity is started on screen rotation (twice).

Please test.

app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesFragment.java
```diff
@@ -112,8 +107,8 @@ public void onActivityCreated(@Nullable Bundle savedInstanceState) {
         super.onActivityCreated(savedInstanceState);
 
         RecyclerView recyclerView = binding.rvAttendeeList;
-        recyclerView.setLayoutManager(new LinearLayoutManager(context));
-        recyclerView.addItemDecoration(new DividerItemDecoration(context, DividerItemDecoration.VERTICAL));
+        recyclerView.setLayoutManager(new LinearLayoutManager(getActivity()));
```
iamareebjamal: Why doing getActivity() everywhere?

app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesPresenter.java
```diff
@@ -41,8 +43,14 @@ public void attach(long eventId, IAttendeesView attendeesView) {
...
-    public void start() {
-        loadAttendees(false);
+    public void start(boolean forceStart) {
+        if (forceStart)
+            progress = 0;
+        if (progress == 0) {
+            loadAttendees(false);
+        } else {
+            attendeesView.showAttendees(attendeeList);
+        }
```
iamareebjamal: Why this?

mejariamol: This is when already data is  loaded in presenter, hence just showing it instead of reloading it from disk again.

mejariamol:
@iamareebjamal Please do test on device. The problem is with attendees fragment. On attendees fragment when screen is rotated for first time everything is fine, no data reload. But when screen is rotated again the activity starts events list activity unexpectedly.

app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesPresenter.java
iamareebjamal: IMO this is the problem of view. In view, we could check that if it is rotated and loading again, we'll just get the attendees, and if it is not, then we'll loaf

mejariamol: >In view, we could check that if it is rotated and loading again

view doesn't have any idea if data is loaded or not. Presenter has. Thats why I am checking it in presenter.

mejariamol:
@iamareebjamal 
problem is - only one object can be retained in the activity using `setRetainInstance(true)`. Hence can't retain both fragment. Looking for workaround. Any suggestions?

iamareebjamal:
@mejariamol Looking for the solution. While that, let's move on to other issues as this is not a critical one. The data is correctly loading in the app and not creating any performance issues either. These kind of things are there due to flawed API of Android core components, even when they are part of the support libraries.

iamareebjamal:
This is the most recommended solution as of now

iamareebjamal:
I think we should close this PR as we'll come up with another solution later

mejariamol:
yeah. closing it. :+1:

PR #220, opened 2017-06-22
feat: Implements attendees filtering 

### LINKED ISSUE #190: Implement attendees search
To be done after #193 
Currently no attendees search present.
Expected:
- search view in app bar on attendees page which can filter attendees by name or email

I am working on it.

mejariamol:
- Adds search view in action bar for attendees fragment
- Implements fuzzy search filtering with minimum fuzzy sort partial ratio 60%

fixes #190 

screenshot_2017-06-23-02-03-01

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesFragment.java
+++ b/app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesFragment.java
@@ -13,15 +13,18 @@ import android.support.v7.widget.DefaultItemAnimator;
 import android.support.v7.widget.DividerItemDecoration;
 import android.support.v7.widget.LinearLayoutManager;
 import android.support.v7.widget.RecyclerView;
+import android.support.v7.widget.SearchView;
 import android.view.LayoutInflater;
+import android.view.Menu;
+import android.view.MenuInflater;
+import android.view.MenuItem;
 import android.view.View;
 import android.view.ViewGroup;
 import android.widget.Toast;
 
 import com.android.databinding.library.baseAdapters.BR;
-import com.mikepenz.fastadapter.FastAdapter;
 import com.mikepenz.fastadapter.adapters.HeaderAdapter;
-import com.mikepenz.fastadapter.adapters.ItemAdapter;
+import com.mikepenz.fastadapter.commons.adapters.FastItemAdapter;
 import com.timehop.stickyheadersrecyclerview.StickyRecyclerHeadersDecoration;
 
 import org.fossasia.openevent.app.OrgaApplication;
@@ -37,6 +40,7 @@ import org.fossasia.openevent.app.event.checkin.AttendeeCheckInFragment;
 import org.fossasia.openevent.app.events.EventListActivity;
 import org.fossasia.openevent.app.qrscan.ScanQRActivity;
 import org.fossasia.openevent.app.utils.Constants;
+import org.fossasia.openevent.app.utils.SearchUtils;
 import org.fossasia.openevent.app.utils.ViewUtils;
 
 import java.util.List;
@@ -48,13 +52,11 @@ import javax.inject.Inject;
  * Use the {@link AttendeesFragment#newInstance} factory method to
  * create an instance of this fragment.
  */
-public class AttendeesFragment extends BaseFragment implements IAttendeesView {
+public class AttendeesFragment extends BaseFragment implements IAttendeesView, SearchView.OnQueryTextListener {
 
     private Context context;
 
     private long eventId;
-
-    private ItemAdapter<Attendee> itemAdapter;
 
     public static final int REQ_CODE = 123;
 
@@ -64,6 +66,7 @@ public class AttendeesFragment extends BaseFragment implements IAttendeesView {
     @Inject
     IAttendeesPresenter attendeesPresenter;
 
+    private FastItemAdapter<Attendee> fastItemAdapter;
     private StickyHeaderAdapter stickyHeaderAdapter;
     private RecyclerView.AdapterDataObserver adapterDataObserver;
     private FragmentAttendeesBinding binding;
@@ -99,11 +102,21 @@ public class AttendeesFragment extends BaseFragment implements IAttendeesView {
             .inject(this);
 
         super.onCreate(savedInstanceState);
+        setHasOptionsMenu(true);
 
         if (getArguments() != null) {
             eventId = getArguments().getLong(EventListActivity.EVENT_KEY);
             attendeesPresenter.attach(eventId, this);
         }
+    }
+
+    public void onCreateOptionsMenu(Menu menu, MenuInflater inflater) {
+        inflater.inflate(R.menu.menu_attendees, menu);
+        MenuItem search = menu.findItem(R.id.search);
+        SearchView searchView = (SearchView) search.getActionView();
+        searchView.setQueryHint(getString(R.string.search_placeholder));
+        searchView.setOnQueryTextListener(this);
+        super.onCreateOptionsMenu(menu, inflater);
     }
 
     @Override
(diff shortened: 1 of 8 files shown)


mejariamol:
@iamareebjamal please review. I am not sure if I can use MVP pattern here.

iamareebjamal:
@mejariamol That's fine because filtering is not the job of presenter now, it's the job of fastadapter, so we don't need to test it

app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesFragment.java
```diff
@@ -97,6 +104,27 @@ public void onCreate(Bundle savedInstanceState) {
     }
 
     @Override
+    public void onCreateOptionsMenu(Menu menu, MenuInflater inflater) {
+        inflater.inflate(R.menu.menu_attendees, menu);
+        MenuItem search = menu.findItem(R.id.search);
+        SearchView searchView = (SearchView) search.getActionView();
+        searchView.setQueryHint("Search Attendees");
```
iamareebjamal: Use strings.xml

app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesFragment.java
```diff
@@ -97,6 +104,27 @@ public void onCreate(Bundle savedInstanceState) {
...
+        searchView.setOnQueryTextListener(new SearchView.OnQueryTextListener() {
```
iamareebjamal: Detach listener on destroy

app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesFragment.java
```diff
@@ -124,6 +152,17 @@ public void onActivityCreated(@Nullable Bundle savedInstanceState) {
 
         fastAdapter.setHasStableIds(true);
         fastAdapter.withEventHook(new AttendeeItemCheckInEvent(this));
+        itemAdapter.withItemFilter(new ItemFilter<>(itemAdapter)
+            .withFilterPredicate(
+                (attendee, query) -> {
+                    int firstNameMatch = FuzzySearch.tokenSortPartialRatio(query.toString().toLowerCase(), attendee.getFirstName().toLowerCase());
+                    if (firstNameMatch > 60) return false;
```
iamareebjamal: Might want to explain here what's happening :stuck_out_tongue: . Maybe create a SearchUtils class if possible

mejariamol: token sort partial ratio returns a value in percentage about similarity between two strings. I tested it with some strings and found that  greater than 60 does well. So we are checking similarity of a query with firstname lastname and email of an attendee if at least one of them is greater than 60.

I will add same in the comment

mejariamol:
@iamareebjamal ready to merge. please review.

app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesFragment.java
iamareebjamal: This can be moved to a general SearchUtil class with a test

app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesFragment.java
```diff
@@ -231,18 +263,29 @@ public void showScanButton(boolean show) {
     public void showAttendees(List<Attendee> attendees) {
         // The list is loaded from presenter, so we just need
         // to notify RecyclerView to update the data
-        itemAdapter.set(attendees);
+        itemAdapter.setNewList(attendees, true);
```
iamareebjamal: Why this?

app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesFragment.java
```diff
@@ -231,18 +263,29 @@ public void showScanButton(boolean show) {
...
         binding.setVariable(BR.attendees, attendees);
         binding.executePendingBindings();
     }
 
     @Override
     public void updateAttendee(int position, Attendee attendee) {
-        itemAdapter.set(position, attendee);
+        itemAdapter.setNewList(attendeesPresenter.getAttendees(), true);
```
iamareebjamal: This throws away the need to calculate the attendee index and recreates the whole layout again

app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesFragment.java
```diff
@@ -231,18 +263,29 @@ public void showScanButton(boolean show) {
...
     }
 
     @Override
     public void showErrorMessage(String error) {
         Toast.makeText(context, error, Toast.LENGTH_SHORT).show();
     }
+
+    @Override
+    public boolean onQueryTextSubmit(String query) {
```
iamareebjamal: Keep these general methods above and View implementation methods below

app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesFragment.java
mejariamol: Its is required to manage correct item update.
Thats where I was working. Just using `set` was giving indexbound exception in filter case.
Hence passing the reference of list again is only option found.
I tested it with 1000+ attendees. Working fine.

app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesFragment.java
mejariamol: Same reason here. Since this has second boolean parameter retainFilter .

app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesFragment.java
iamareebjamal: If one attendee is updated, is the whole view updated or just this one attendee? You can visually see the difference. If the whole view is updated, better use DiffUtils, it does look file with 1000 attendees, but is not the matter of numbers, just this that there shouldn't be a need to update each view show to register change in one

mejariamol: Will create separate issue for implementing diffutils.  This will do good for now. :+1:

mejariamol:
@iamareebjamal @rohanagarwal94 @Shailesh351 @arp95 please review

iamareebjamal:
@mejariamol After the filter is cleared, check in status in attendee list does not remain reactive, meaning that the it is not reflected correctly

mejariamol:
@iamareebjamal checking.

mejariamol:
@iamareebjamal problem comes in toggling check in from filtered list. Couldn't find workaround in fast adapter.

iamareebjamal:
I'll take a look later. The thing is that you pass the copied list to fast adapter and swap the attendee and apply the filter again. This is the correct workflow

mejariamol:
@iamareebjamal if you are adding this feature for screen cast, please show check in from direct list only.

mejariamol:
>The thing is that you pass the copied list to fast adapter and swap the attendee and apply the filter again

problem is with reverse binding

iamareebjamal:
The video is getting long so just showing merged features. Only basic features are taking long enough. Scanning of QR is a major feature, thus, requires some attention

mejariamol:
@iamareebjamal please do test and review. :)

mejariamol:
@iamareebjamal please do test on device and review the code. :-)

mejariamol:
@iamareebjamal please review.

app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesFragment.java
```diff
@@ -215,9 +247,8 @@ public void showScanButton(boolean show) {
 
     @Override
     public void showAttendees(List<Attendee> attendees) {
-        // The list is loaded from presenter, so we just need
-        // to notify RecyclerView to update the data
-        itemAdapter.set(attendees);
+        fastItemAdapter.clear();
+        fastItemAdapter.add(attendees);
```
mejariamol: Okay. I will check. :+1:

mejariamol: done :+1:

app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesPresenter.java
```diff
@@ -112,7 +112,6 @@ private void listenToModelChanges() {
 
         attendeeListener.getNotifier()
             .filter(attendeeModelChange -> attendeeModelChange.getAction().equals(BaseModel.Action.UPDATE))
-            .distinctUntilChanged()
```
mejariamol: Using `distinctUntilChanged` filter out subsequent check ins update after first check in, irrespective of identity.
Why do we need it. Does our listener sends same modelChange more than once ?

iamareebjamal: Yes, the current code is faulty, it'll filter the updates, that's why I asked to add distinctUntilChanged after the map operation. Then it should work fine.

The problem is not of sending the same model again and again, but about the granular changes which will not affect the UI causing it to trigger update again and again.

mejariamol: but adding `distinctUntilChanged` is causing issue in updates. Without it, it is working fine. Please do test on device.

mejariamol: the question is how `distinctUntilChanged` is going to compare model? Don't we have to set that logic?

iamareebjamal: OK, remove for now. We are not making any granular changes :+1:

mejariamol: It is removed in the latest commit.

app/src/main/java/org/fossasia/openevent/app/utils/SearchUtils.java
```diff
@@ -0,0 +1,28 @@
...
+     * Token sort partial ratio returns a value in percentage about similarity
+     * between two strings. The same is tested and found that
+     * greater than 60 does well. So we are checking similarity of a query with
+     * first name, last name and email of an attendee if at least one of them is greater than 60.
+     *
+     * @param attendee to be filtered
+     * @param query search query
+     * @return boolean saying filter or not
+     */
+    public static boolean shallFilterAttendee(Attendee attendee, CharSequence query) {
```
iamareebjamal: Make this a generic method, not just for an attendee. First argument may be charsequence and second one String varargs

app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesFragment.java
iamareebjamal: Keep these general methods above and View implementation methods below

app/src/main/java/org/fossasia/openevent/app/event/attendees/AttendeesFragment.java
```diff
@@ -215,9 +227,9 @@ public void showScanButton(boolean show) {
 
     @Override
     public void showAttendees(List<Attendee> attendees) {
-        // The list is loaded from presenter, so we just need
-        // to notify RecyclerView to update the data
-        itemAdapter.set(attendees);
+        //fastItemAdapter.clear();
+        //fastItemAdapter.add(attendees);
```
iamareebjamal: Remove these comments please

mejariamol: please review now.

PR #221, opened 2017-06-23
feat: Group events into live, past and upcoming

### LINKED ISSUE #152: Grouping events in the event list
To be done after #156 

Currently all events are shown in an event list.

**Expected:** Events should be grouped under live, past and future on basis of event start-end time.

mejariamol:
- Implements comparable interface on event model
    - sorts events in order of live, upcoming and past
    - for both past, lately ended will be before in list
    - for both live, lately started will be before in list
    - for both upcoming earliest will be before
- Implements sticky header adapter on eventslistadapter
- Modifies event layout

fixes #152

screenshot_2017-06-23-18-13-57

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/app/events/EventsListAdapter.java
+++ b/app/src/main/java/org/fossasia/openevent/app/events/EventsListAdapter.java
@@ -25,9 +25,13 @@ import java.util.List;
 
 class EventsListAdapter extends RecyclerView.Adapter<EventsListAdapter.EventRecyclerViewHolder> implements StickyRecyclerHeadersAdapter<EventsHeaderViewHolder>{
 
-    private static final long LIVE_EVENT = 1;
-    private final long PAST_EVENT = 2;
-    private static final long UPCOMING_EVENT = 3;
+    private static final int LIVE_EVENT = 1;
+    private static final int PAST_EVENT = 2;
+    private static final int UPCOMING_EVENT = 3;
+
+    private static final String HEADER_LIVE = "LIVE";
+    private static final String HEADER_PAST = "PAST";
+    private static final String HEADER_UPCOMING = "UPCOMING";
 
     private List<Event> events;
     private Context context;
@@ -75,21 +79,8 @@ class EventsListAdapter extends RecyclerView.Adapter<EventsListAdapter.EventRecy
     @Override
     public long getHeaderId(int position) {
         Event event = events.get(position);
-        DateUtils dateUtils = new DateUtils();
         try {
-            Date startDate = dateUtils.parse(event.getStartTime());
-            Date endDate = dateUtils.parse(event.getEndTime());
-            Date now = new Date();
-            if (now.after(startDate)) {
-                if (now.before(endDate)) {
-                    return LIVE_EVENT;
-                } else {
-                    return PAST_EVENT;
-                }
-            } else {
-                return UPCOMING_EVENT;
-            }
-
+            return getEventStatus(event);
         } catch (ParseException e) {
             e.printStackTrace();
         }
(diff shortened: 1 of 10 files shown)


mejariamol:

iamareebjamal:
I think showing date will look better as subtitle below title

iamareebjamal:
@mejariamol Date functions are repeated a lot. Create a date util class and also compare util if the code for comparison is also shared

mejariamol:
>Date functions are repeated a lot. Create a date util class

But it will include creating `DateFormat` class and setting timezone to UTC on each time it is used. With present approach `DateFormat` is created once and used at that spot.

iamareebjamal:
You can create a static DateFormat with UTC

iamareebjamal:
BTW, there is no need to set UTC, just show the date in Event timezone, not UTC

mejariamol:
It is converted into UTC for comparison

mejariamol:
@iamareebjamal Some tests are failing due to `java.text.DateFormat` which is used for parsing date. It is not thread safe.
>Date formats are not synchronized. It is recommended to create separate format instances for each thread. If multiple threads access a format concurrently, it must be synchronized externally.

from java doc. Hence it is doing job in the app as we are sorting in async but failing tests. 

shall we pass the tests for this?

iamareebjamal:
The tests are failing due to local time being different in your settings and CI settings. We can leave the tests for now, but create an issue for the same. And that's why I asked to move the date logic out of presenter so that we can mock it in the tests

mejariamol:
Ok. Will try using Joda-Time instead which is more clearer.

iamareebjamal:
Nope, that library is huge and causes severe performance issues on Android! This is a 15 liner solution

Take a look at DateUtils in open android project

mejariamol:
Actually with present code everything is working fine in the app.
please review for other things.
I am not getting why test terminate when comes at compare function where `DatetimeFormat` is initiated.
I tried on Joda-time , same behaviour , working for app but failing test.

Will create issue for the same.

mejariamol:
you may run the code if got time. Please see if I am missing something.

mejariamol:
@iamareebjamal what to do with this? It's working fine in the app. Only problem is with tests.

iamareebjamal:
Will take a look at it in about an hour

iamareebjamal:
@mejariamol Please look at the changes I made in last commit and test if everything's working correctly

mejariamol:
@iamareebjamal yeah. working fine. Thanks :)

mejariamol:
@arp95 @rohanagarwal94 @Shailesh351 guys, please review

mejariamol:
@iamareebjamal please review. Tests were failing due to IEventRepository's name mismatch. Corrected.

iamareebjamal:
They were being fixed by #233

mejariamol:
No problem. I have just corrected IEventRepository's name. :+1:

mejariamol:
@iamareebjamal tests from SimpleModelTest are failing . Please take a look.

iamareebjamal:
Working fine locally. Seems internal issue with travis or roboelectric. Looking for workaround

iamareebjamal:
@mejariamol Will be fixed by #233

app/src/main/java/org/fossasia/openevent/app/events/EventsListAdapter.java
```diff
@@ -62,6 +73,62 @@ void showInitialEvent() {
...
+    }
+
+    @Override
+    public EventsHeaderViewHolder onCreateHeaderViewHolder(ViewGroup viewGroup) {
+        return new EventsHeaderViewHolder(EventSubheaderLayoutBinding.inflate(LayoutInflater.from(viewGroup.getContext()), viewGroup, false));
+    }
+
+    @Override
+    public void onBindHeaderViewHolder(EventsHeaderViewHolder holder, int position) {
+        final String LIVE = "LIVE";
```
iamareebjamal: Use the constants declared above

app/src/main/java/org/fossasia/openevent/app/events/EventsListAdapter.java
```diff
@@ -62,6 +73,62 @@ void showInitialEvent() {
...
+                    holder.bindHeader(PAST);
+                }
+            } else {
+                holder.bindHeader(UPCOMING);
+            }
+
+        } catch (ParseException e) {
+            e.printStackTrace();
+        }
+    }
```
iamareebjamal: This block is repeated. Create a function

mejariamol:
@iamareebjamal done.

PR #223, opened 2017-06-23
fix: Correct empty view display

### LINKED ISSUE #222: Empty layout appears at start itself
Currently on any data loading page, empty layout is shown from start itself with progress bar.

Expected: On page start , only progress bar should appear with blank background. Then on data load failure empty layout should appear.

mejariamol:
- Adds showEmptyView method in interface
- Hides empty view on page start
- Hides empty view on swipe refresh
- Modifies empty view layout

fixes #222

img-20170624-wa0002

### DIFF
--- a/app/src/test/java/org/fossasia/openevent/app/presenter/AttendeePresenterTest.java
+++ b/app/src/test/java/org/fossasia/openevent/app/presenter/AttendeePresenterTest.java
@@ -16,6 +16,7 @@ import org.mockito.Mockito;
 import org.mockito.junit.MockitoJUnit;
 import org.mockito.junit.MockitoRule;
 
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 
@@ -59,7 +60,6 @@ public class AttendeePresenterTest {
     public void setUp() {
         attendeesPresenter = new AttendeesPresenter(eventModel);
         attendeesPresenter.attach(id, attendeesView);
-        attendeesPresenter.setAttendeeList(attendees);
 
         RxJavaPlugins.setComputationSchedulerHandler(scheduler -> Schedulers.trampoline());
         RxAndroidPlugins.setInitMainThreadSchedulerHandler(schedulerCallable -> Schedulers.trampoline());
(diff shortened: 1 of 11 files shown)


mejariamol:

iamareebjamal:
Um.. You could just set default visibility to GONE

mejariamol:
But also had to handle after swipe refresh.

iamareebjamal:
data binding should handle it automatically. When it is possible through data binding, there's no point in adding complexity in presenter

mejariamol:
yes, but just knowing events size view can't decide whether to hide or show empty view

mejariamol:
For example even for items size 0, on swipe refresh , empty view should hide and only progress bar should appear and on complete it should appear if items size 0. 
Hence its just like implementing progress bar. So I think better to be  implemented in presenter view

iamareebjamal:
This is adding additional complexity to otherwise very basic idea. It's not actually bad to show empty view while loading because there is nothing present right now. Besides, if you add this interaction, you have to check it on errors too, that the empty layout is shown again or not. Plus, remember that we'll be moving the progress bar itself from presenter to data binding by adding a field in model, so this step is actually going backwards. Once we add that field in model for describing transient state, we'll just add a boolean check to see that if it is transient, hide the empty view and if it's not, display it. But now if we move logic in presenter, we'll have to move it back to layout some time ahead

mejariamol:
@iamareebjamal 
>we'll be moving the progress bar itself from presenter to data binding by adding a field in model

Yes, thats why let keep empty view visibility logic in presenter for now. both shall be moved to data binding together.

One thing is clear that we need this behaviour - 
>Adds showEmptyView method in interface
Hides empty view on page start
Hides empty view on swipe refres

So I think lets keep this logic in presenter for now. Will create issue for handling progress bar and empty view visibility using databinding.

iamareebjamal:
Then please test that it works in all cases

mejariamol:
I have tested on device. Will add test in presenters by morning. :)

mejariamol:
@iamareebjamal  added tests. please review.

app/src/test/java/org/fossasia/openevent/app/presenter/AttendeePresenterTest.java
```diff
@@ -122,10 +125,12 @@ public void shouldLoadAttendeesSuccessfully() {
         attendeesPresenter.loadAttendees(false);
 
         inOrder.verify(attendeesView).showProgressBar(true);
+        inOrder.verify(attendeesView).showEmptyView(false);
```
iamareebjamal: Remove the false cases and add them as separate tests. These tests have become very self dependent. Each test should handle only one aspect of the view.

Will simplify them later as this is a low priority issue

iamareebjamal: And add a case where error happens while updating

mejariamol:
@iamareebjamal done :+1:

iamareebjamal:
@arp95 @rohanagarwal94 @Shailesh351 guys, please review

mejariamol:
@iamareebjamal done :+1:

mejariamol:
@arp95 @rohanagarwal94 @Shailesh351 guys, please review

PR #256, opened 2017-07-03
fix: updating local database on force reload

### LINKED ISSUE #255: Old data is not removed on swipe refresh success
On swipe refresh new data is saved.(Updated)

**Expected:** On swipe refresh success complete concerned table must be cleared before saving new data. So that this will ensure that data deleted on the server never stays locally.

mejariamol:
- clears table (Attendee, Event) before saving new data on force reload

fixes #255

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/app/data/models/Ticket.java
+++ b/app/src/main/java/org/fossasia/openevent/app/data/models/Ticket.java
@@ -3,6 +3,7 @@ package org.fossasia.openevent.app.data.models;
 import android.support.annotation.VisibleForTesting;
 
 import com.raizlabs.android.dbflow.annotation.ForeignKey;
+import com.raizlabs.android.dbflow.annotation.ForeignKeyAction;
 import com.raizlabs.android.dbflow.annotation.PrimaryKey;
 import com.raizlabs.android.dbflow.annotation.Table;
 import com.raizlabs.android.dbflow.structure.BaseModel;
@@ -21,7 +22,7 @@ public class Ticket extends BaseModel {
     private long quantity;
     private String type;
 
-    @ForeignKey(stubbedRelationship = true)
+    @ForeignKey(stubbedRelationship = true, onDelete = ForeignKeyAction.CASCADE)
     private Event event;
 
     public Ticket() {}
--- a/app/src/main/java/org/fossasia/openevent/app/data/EventRepository.java
+++ b/app/src/main/java/org/fossasia/openevent/app/data/EventRepository.java
@@ -112,7 +112,8 @@ public class EventRepository implements IEventRepository {
 
         Observable<Event> networkObservable = Observable.defer(() ->
             eventService.getEvents(getAuthorization())
-                .doOnNext(events -> databaseRepository.saveList(Event.class, events)
+                .doOnNext(events -> databaseRepository.deleteAll(Event.class)
+                    .concatWith(databaseRepository.saveList(Event.class, events))
                     .subscribe())
                 .flatMapIterable(events -> events))
                 .doOnEach(eventNotification -> {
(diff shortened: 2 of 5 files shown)


app/src/main/java/org/fossasia/openevent/app/data/models/Ticket.java
```diff
@@ -19,7 +20,7 @@
     private long quantity;
     private String type;
 
-    @ForeignKey(stubbedRelationship = true)
+    @ForeignKey(stubbedRelationship = true, onDelete = ForeignKeyAction.CASCADE)
```
iamareebjamal: This will delete the event is ticket is deleted

mejariamol: do you mean - it will delete event **if** ticket is deleted ?

iamareebjamal: Yes

mejariamol: So what can be done to just delete ticket and social link on event deletion ?

iamareebjamal: The onDelete trigger in the Event itself will need to be changed, but I have already set it to CASCADE. There is a bug however which does not cascade deletion in the library itself

Now, it won't create any problem because these are linked with the ID of event, so won't be loaded when event is deleted

So, deleting the event only should be enough

mejariamol: no but just deleting events is giving error of foreign key constraint.

iamareebjamal: Adding this is fixing it?

iamareebjamal: Actually just deleting event is giving error because the attendees and tickets are also linked to event.

mejariamol: yeah. we want to set it up such that tickets and social link will delete on event deletion not the other way. any idea how we can do that?

iamareebjamal: Still haven't got a reply from the developer about the bug fix, for now, we'll have to manually remove all  the events and their linked fields, I'll look for a solution tomorrow

mejariamol: >This will delete the event if ticket is deleted

Why will it delete event if ticket is deleted. Here ticket is child and event is parent. so the action is applied to the child from parent not the other way. Here am I missing something ?

After setting onDelete action, I was able to use `deleteAll` on Event.class.

I didn't actually get your issue https://github.com/Raizlabs/DBFlow/issues/1332. please explain.

iamareebjamal: on delete is a trigger which works on foreign keys. Meaning if this record containing this foreign key is deleted, what do you want me to do with the foreign key model, CASCADE means delete that too. Meaning, it is saying if the event foreign key is deleted (when ticket will be deleted), delete the event too. You can actually try by calling ticket.delete(). 

The bug is that this constraint is only satisfied if you call model.delete(), not the model adaper.delete(model)

Deleting event will work fine obviously, test that is deleting a ticket also is deleting event or not. But, it won't delete now because of this bug, and cause trouble later when it is fixed.

iamareebjamal: https://stackoverflow.com/questions/2914936/mysql-foreign-key-constraints-cascade-delete

mejariamol: >Meaning, it is saying if the event foreign key is deleted (when ticket will be deleted), delete the event too

No it works in the other way. 
>Foreign key ON DELETE and ON UPDATE clauses are used to configure actions that take place when deleting rows from the parent table (ON DELETE), or modifying the parent key values of existing rows (ON UPDATE).

here parent is event
and child is ticket and social link.

So I think nothing wrong. What do you say?

iamareebjamal: I guess I would have got it wrong if it is working. Please continue. I thought it worked through references

mejariamol:
@iamareebjamal have tested on device, working fine. Please review.

PR #260, opened 2017-07-05
fix: clearing database on different user login

### LINKED ISSUE #252: local data is not cleared on different user login
Currently we are matching username (email id) of new user to decide if local data is to be removed.

But as we are using (and shall be using) separate server for _dev_ branch where email can be same hence previous data stored is not removed in that case.

Expected: username and userid both shall be checked to decide previous data removal.

@iamareebjamal what is your opinion?

mejariamol:
- method `LoginModel::login` is modified to just get access token
- new method `LoginModel::isPreviousUser` is added to check different user login
- method `LoginPresenter::login` is modified accordingly to clear database on different user login and starts data loading after that only.

fixes #252

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/app/data/LoginModel.java
+++ b/app/src/main/java/org/fossasia/openevent/app/data/LoginModel.java
@@ -1,12 +1,14 @@
 package org.fossasia.openevent.app.data;
 
+import android.support.annotation.VisibleForTesting;
+
 import org.fossasia.openevent.app.data.contract.ILoginModel;
 import org.fossasia.openevent.app.data.contract.IUtilModel;
 import org.fossasia.openevent.app.data.db.contract.IDatabaseRepository;
 import org.fossasia.openevent.app.data.models.Login;
-import org.fossasia.openevent.app.data.models.LoginResponse;
 import org.fossasia.openevent.app.data.models.User;
 import org.fossasia.openevent.app.data.network.EventService;
+import org.fossasia.openevent.app.data.repository.contract.IEventRepository;
 import org.fossasia.openevent.app.main.MainActivity;
 import org.fossasia.openevent.app.utils.Constants;
 import org.fossasia.openevent.app.utils.JWTUtils;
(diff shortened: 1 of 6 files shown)


mejariamol:
@iamareebjamal please test on device and review. Will be adding test later.

app/src/main/java/org/fossasia/openevent/app/data/LoginModel.java
```diff
@@ -43,18 +48,20 @@ public LoginModel(IUtilModel utilModel, EventService eventService, IDatabaseRepo
...
-                            .subscribe()
-                    ));
+                .observeOn(AndroidSchedulers.mainThread());
+    }
+
+    @Override
+    public Observable<Boolean> isPreviousUser() {
+        return Observable.combineLatest(eventService.getUser(getAuthorization()),
+            databaseRepository.getAllItems(User.class).switchIfEmpty(Observable.just(new User(""))).firstElement().toObservable(), this::areSameUsers)
+            .subscribeOn(Schedulers.io())
```
iamareebjamal: Why fetching the user from EventService here. It should be loaded at login itself. I mean to say that presenter should not trigger the logic to handle the previous user. Only LoginModel should do it. The presenter should only get the login success or error

app/src/main/java/org/fossasia/openevent/app/data/LoginModel.java
```diff
@@ -43,18 +48,20 @@ public LoginModel(IUtilModel utilModel, EventService eventService, IDatabaseRepo
...
+            .observeOn(AndroidSchedulers.mainThread());
+    }
+
+    private Boolean areSameUsers(User user1, User user2) {
+        return (user1.getId() == user2.getId() && user1.getEmail().equals(user2.getEmail()));
```
iamareebjamal: Override equals in User class instead

app/src/main/java/org/fossasia/openevent/app/data/LoginModel.java
iamareebjamal: Also, instead of using combineLatest with switchIfEmpty. First check if the database entry of users is empty and if it is, then no need to apply check at all. If it is not, then apply chaining logic

mejariamol: @iamareebjamal Yeah. Initially I tried doing all the stuff in login itself but I couldn't manage synchronisation between database deletion -> actual login( starting MainActivity) . Please suggest an approach.

iamareebjamal: OK. Will do soon :+1:

mejariamol: @iamareebjamal hey I think it would better to keep them separate. `LoginModel::login` should just handle login ie getting access token. Then before starting main activity it is presenter's work to check for data consistency and take actions accordingly. So I think current approach is fine. What do you say? doing all the stuff in the login itself will increase complexity.

iamareebjamal: The thing is presenter should never have access to db, it's not presenter's work. That's what models are there for. Presenter's job is to only get information from the model, without caring from where it is getting from, and process it to display it to view. All actions taken on data are the work of model

iamareebjamal: Models are actually the most complex part of MVP, as they are self-managed, they don't need input from presenter to trigger something

mejariamol:
@iamareebjamal please review this.

iamareebjamal:
@mejariamol There are no changes from the last time I reviewed it

mejariamol:
>@iamareebjamal Yeah. Initially I tried doing all the stuff in login itself but I couldn't manage synchronisation between database deletion -> actual login( starting MainActivity) . Please suggest an approach.

>OK. Will do soon :+1:

I thought you were going to suggest another approach.

app/src/main/java/org/fossasia/openevent/app/data/LoginModel.java
mejariamol: thing is that login view should start main activity only after completion of database handling. These two processes should be in sequence not concurrently.

iamareebjamal: This will be done in sequence only

mejariamol: but if database handling is done in onNext and a user is returned to the login presenter where it will start main activity as it gets the user.

iamareebjamal: Let me try it

iamareebjamal:
@mejariamol Please check and rewrite tests

app/src/main/java/org/fossasia/openevent/app/data/LoginModel.java
```diff
@@ -1,67 +1,81 @@
...
             .subscribeOn(Schedulers.io())
             .observeOn(AndroidSchedulers.mainThread());
     }
 
-    private Boolean areSameUsers(User user1, User user2) {
-        return (user1.getId() == user2.getId() && user1.getEmail().equals(user2.getEmail()));
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    public Single<Boolean> isPreviousUser() {
+        return databaseRepository.getAllItems(User.class)
+            .first(EMPTY)
```
mejariamol: was this suppose to be `firstOrDefault` ? @iamareebjamal

iamareebjamal: No

mejariamol:
@iamareebjamal please review

iamareebjamal:
@mejariamol I must be really out of mind to forget that identity of the user is present in JWT token. Fix the tests, we'll merge this and I'll add parsing code and tests and send a separate PR

mejariamol:
@iamareebjamal please review. :+1:

iamareebjamal:
@Shailesh351 @arp95 @shreyasnbhat @rohanagarwal94 Please review

Issue #274, opened 2017-07-06
Data loading takes longer on Events and Attendees page


mejariamol:
sub issue of #273 

Currently the pages fails to load data sometimes and keep on showing progress bar for long. 
The reason behind misbehaviour of Computation scheduler in data loading observables. Removing it solves issue for now.
**Future action:** implement sorting in fastAdapter itself. Creating issue for the same. #275  But for now just removing computation scheduler.

iamareebjamal:
This was fixed by #277

PR #277, opened 2017-07-06
fix: Remove blocking call to fix synchronization errors

### LINKED ISSUE #273: Handle Synchronisation errors
Somehow synchronisation errors have crept in the app which are only happening in certain cases. Currently, debugging and fixing them

- [x] Attendees page subscribe not happening
- [x] Events page subscribe not happening
- [x] Ticket analyser subscribe not happening

iamareebjamal:
Fixes #273 and #274 

- Fix blocking call on computation scheduler


iamareebjamal:
@arp95 @Shailesh351 @rohanagarwal94 Please review

PR #279, opened 2017-07-06
feat: Add DateUtils and DateService with tests

### LINKED ISSUE #264: Add DateUtils and delegate all date related tasks to this class
Add DateUtils and delegate all date-related tasks to this class from presenters and models

Taking this up

iamareebjamal:
Fixes #264 

- Add DateUtils and DateService
- Add tests
- Move all date related handling in DateService and DateUtils

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/app/utils/DateService.java
+++ b/app/src/main/java/org/fossasia/openevent/app/utils/DateService.java
@@ -1,0 +1,53 @@
+package org.fossasia.openevent.app.utils;
+
+import org.fossasia.openevent.app.data.models.Event;
+
+import java.text.ParseException;
+import java.util.Date;
+
+public class DateService {
+
+    private static final String LIVE_EVENT = "LIVE";
+    private static final String PAST_EVENT = "PAST";
+    private static final String UPCOMING_EVENT = "UPCOMING";
+
+    public static int compareEventDates(Event one, Event two) {
+        Date now = new Date();
+        try {
+            Date startDate = DateUtils.getDate(one.getStartTime());
+            Date endDate = DateUtils.getDate(one.getEndTime());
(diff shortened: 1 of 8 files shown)


app/src/main/java/org/fossasia/openevent/app/utils/DateService.java
```diff
@@ -0,0 +1,53 @@
+package org.fossasia.openevent.app.utils;
+
+import org.fossasia.openevent.app.data.models.Event;
+
+import java.text.ParseException;
+import java.util.Date;
+
+public class DateService {
```
mejariamol: Better we keep this (comparison and getting event status) in `Event.class` itself. That will be easier to manage as this class will become messy on further additions.

iamareebjamal: It's not true. Larger the class gets, the messier it is. See the composition of standard libraries in Java, almost all of them have just method names in their class and everything is delegated to specific objects.

https://github.com/ReactiveX/RxJava/blob/2.x/src/main/java/io/reactivex/Observable.java

And this is not something out of the blue I am advocating, this is Facade Design Pattern

https://en.wikipedia.org/wiki/Facade_pattern

Distributing code among classes is opposite of making things messy. As this is a model class, it already pretty huge, or else, there should be a simple rule, break the code if it exceeds 200 lines

mejariamol: >Distributing code among classes is opposite of making things messy

I was just suggesting to keep these two methods in event model as these are related to event model. No problem. :+1:

mejariamol: @arp95 @Shailesh351 @rohanagarwal94 Please review

PR #285, opened 2017-07-07
feat: Check attendee check in indicator

### LINKED ISSUE #263: Replace checked in status in Attendee Layout to something simpler
Currently, UI shows a long text saying if attendee is checked in or not, this is a waste of space and should be simplified

### LINKED ISSUE #284: SearchView detaches even if not attached
Sometimes, onDetach is called before the menu was able to initialize and hence the call of setOnQueryChangeListener causes NPE

I'm sending a PR

iamareebjamal:
Fixes #263

- Add check and cross icon instead of text

device-2017-07-07-201031


iamareebjamal:
@mejariamol I think it'll look OK when ticket info will be shown on it. What do you think?

PR #286, opened 2017-07-07
fix: Add null check on searchView

### LINKED ISSUE #284: SearchView detaches even if not attached

iamareebjamal:
Fixes #284 

- Add null check on searchView


mariobehling:
@iamareebjamal Please ping other devs for review so we can ensure a timely review of at least two devs and keep the fast pace.

Issue #267, opened 2017-07-06
Add Tickets page to show the types of tickets


iamareebjamal:
- Add Tickets page to show each ticket types
- Add ability to add a ticket type

To be done using new API

iamareebjamal:
To be taken after #301 is merged

PR #287, opened 2017-07-08
feat: Add ticket chart analytics

### LINKED ISSUE #261: Show graph indicating ticket sales
Show graph indicating ticket sales on daily basis

Taking this up

iamareebjamal:
Fixes #261

- Add chart view for ticket analytics
- Add chart mutator for handling analytics logic

device-2017-07-09-000009

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -40,6 +40,14 @@ android {
     dataBinding {
         enabled true
     }
+    defaultConfig {
+        multiDexEnabled true
+    }
+    dexOptions {
+        jumboMode true
+        incremental true
+        javaMaxHeapSize "4g"
+    }
     compileOptions {
         sourceCompatibility JavaVersion.VERSION_1_8
         targetCompatibility JavaVersion.VERSION_1_8
@@ -106,6 +114,7 @@ dependencies {
     compile 'com.github.Ilhasoft:data-binding-validator:0.6.4'
     compile 'io.sentry:sentry-android:1.3.0'
     provided 'org.projectlombok:lombok:1.16.18'
+    compile 'com.github.PhilJay:MPAndroidChart:v3.0.2'
     debugCompile 'com.squareup.leakcanary:leakcanary-android:1.5.1'
     releaseCompile 'com.squareup.leakcanary:leakcanary-android-no-op:1.5.1'
     annotationProcessor 'com.jakewharton:butterknife-compiler:8.5.1'
--- a/app/proguard-rules.pro
+++ b/app/proguard-rules.pro
@@ -51,3 +51,5 @@
 # General
 -keepattributes SourceFile,LineNumberTable,*Annotation*,EnclosingMethod,Signature,Exceptions,InnerClasses
 -keep class android.support.v7.widget.SearchView { *; }
+
+-keep class com.github.mikephil.charting.** { *; }
--- a/app/src/main/java/org/fossasia/openevent/app/event/detail/ChartAnalyser.java
+++ b/app/src/main/java/org/fossasia/openevent/app/event/detail/ChartAnalyser.java
@@ -1,0 +1,182 @@
+package org.fossasia.openevent.app.event.detail;
+
+import android.graphics.Color;
+import android.graphics.Paint;
+import android.support.annotation.ColorInt;
+import android.support.annotation.ColorRes;
+
+import com.github.mikephil.charting.charts.LineChart;
+import com.github.mikephil.charting.components.YAxis;
+import com.github.mikephil.charting.data.Entry;
+import com.github.mikephil.charting.data.LineData;
+import com.github.mikephil.charting.data.LineDataSet;
+import com.github.mikephil.charting.utils.EntryXComparator;
+
+import org.fossasia.openevent.app.R;
+import org.fossasia.openevent.app.data.contract.IUtilModel;
+import org.fossasia.openevent.app.data.repository.AttendeeRepository;
+import org.fossasia.openevent.app.utils.DateUtils;
+
+import java.text.ParseException;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import javax.inject.Inject;
+
+import io.reactivex.Completable;
+
+import static org.fossasia.openevent.app.event.detail.TicketAnalyser.TICKET_DONATION;
+import static org.fossasia.openevent.app.event.detail.TicketAnalyser.TICKET_FREE;
+import static org.fossasia.openevent.app.event.detail.TicketAnalyser.TICKET_PAID;
+
+public class ChartAnalyser {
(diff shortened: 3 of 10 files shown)


iamareebjamal:
This is a mock representation. Need to discuss the analysis implementation @mejariamol

mejariamol:
@iamareebjamal first will have to decide what data will be shown in the chart. that is what will be on y axis.

iamareebjamal:
This is per day ticket sale analytics. Y axis is amount of tickets bought and X axis is the date when ticket was bought

mejariamol:
I think we should have a separate chart for paid tickets with total earn on the Y axis. What do you say? As for paid tickets amount will be more informative that just a number. @iamareebjamal

iamareebjamal:
That will be implemented in Tickets Page #267

mejariamol:
Okay. So for implementation of sales versus time, we will need a sorted Ticket list with time of purchase

It can be achieved by sorting Attendee list with `Order` using `completedAt` attribute.  and creating sorted ticket list by iterating through these sorted attendees.

iamareebjamal:
That was what I was thinking. Instead of iterating through tickets, we should iterate through orders. Because orders are less, and are done on a single date. What do you think

mejariamol:
We will need:
1. ticket type( from Ticket)
2. completed at (from order)
3. quantity (from order)

mejariamol:
Just have to think how can we get this info sorted with `completedAt`

iamareebjamal:
We don't need to sort at all, this will be handled by the charting library

mejariamol:
Ok, so what values does charting library take?

iamareebjamal:
It takes a point (X, Y) which in our case will be (purchased, date)

mejariamol:
but order and ticket are related by an attendee. How can we get ticket type by iterating through orders?

iamareebjamal:
order will have one to many relationship with attendees

mejariamol:
Yeah. then it is better to iterate through orders.

mejariamol:
Once we should check complexities for both approaches

mejariamol:
Orders = m
attendees = n
iterating through to orders will have m X n
but through attendees it will have just n

Is is correct? :smile:

iamareebjamal:
No, because we have to get the order from attendees as well, and orders will be repeated, so 20 attendees will have the same order

And if we parse by order, then total attendees will be parsed only for their count and type, but it'll be done in 1 pass only

app/src/main/java/org/fossasia/openevent/app/event/detail/ChartMutator.java
```diff
@@ -0,0 +1,73 @@
...
+    private LineSet paidSet = new LineSet();
+    private LineSet donationSet = new LineSet();
+
+    protected ChartMutator(IUtilModel utilModel) {
+        this.utilModel = utilModel;
+
+        gridPaint.setColor(Color.parseColor("#992ecc71"));
+        gridPaint.setStrokeWidth(5);
+
+        freeSet = new LineSet(new String[] {"Jan", "Feb", "Mar"}, new float[] {1.2f, 3.6f, 4.5f});
```
mejariamol: Seems like we will have to calculate month wise tickets separately. ?

iamareebjamal: This is just mock data, we won't do month wise analysis, we will do aggregate analysis of total sales over the time

mejariamol: Ok. So what will be our time interval?

iamareebjamal: No time interval, just date by sales mapping, it'll be automatically handles by the graphing library

iamareebjamal:
device-2017-07-19-095615

@mejariamol Please review

iamareebjamal:
@mejariamol Can you please check now. I accidentally used Java 8 forEach construct

mejariamol:
@iamareebjamal Some suggestions:
- please use same colors in the chart and for progress bars.
- Currently, it is not clear what actually the chart shows. Please add required info at axes.
- please take care of extremity:
  - what will be shown when no tickets are sold

iamareebjamal:
Same colors are used in both cases. 

Axes will not give an immersive experience, and tooltips will handle the data information. The chart is a standard sales/time which is quite known form of analytic chart.

About extremities, I don't think the app will crash, just show an empty chart. I will check it

mejariamol:
>Axes will not give an immersive experience, and tooltips will handle the data information. The chart is a standard sales/time which is quite known form of analytic chart.

but at least it should show values at axes. :smile:

iamareebjamal:
@mejariamol Imagine an event with 1000 tickets. How will the axes look then? We don't have logic to dynamically show the appropriate amount of labels on the axis. Probably in future

About the colors, I have used primary colors for lines and for dots, I have used the light color. The shade is used in progress bars to show complete incomplete data. What else should be done here?

mejariamol:
>We don't have logic to dynamically show the appropriate amount of labels on the axis.

For now, let's keep the number of labels at axes constant. Their values will be assigned dynamically. For example, 4 labels on the Y axis and 3 labels on the X axis. 
Please try if we can set logic to assign these 7 variables dynamically. Because showing chart without any value at axes is really pointless.

mejariamol:
@iamareebjamal is it possible now to switch chart library to [PhilJay/MPAndroidChart](https://github.com/PhilJay/MPAndroidChart). It is well documented. what do you say? I shall help in switching.

iamareebjamal:
@mejariamol I don't know what such features it brings which WilliamCharts doesn't. I recommend you to check out its sample app https://play.google.com/store/apps/details?id=com.db.williamchartdemo&hl=en

iamareebjamal:
And about the axes. I am not saying that we shouldn't add it. It was never my implication. The chart is far from complete, but if we set out to make it perfect in this PR only, it'll take forever. Issues should be such that they are solved in a day. This already took a lot of time. There are several things remaining on it and should be solved as other issues, so that we can independently and iteratively develop it.

mejariamol:
>I don't know what such features it brings which WilliamCharts doesn't. I recommend you to check out its sample app https://play.google.com/store/apps/details?id=com.db.williamchartdemo&hl=en

yeah. right. WilliamCharts also provides good doc. :+1:

mejariamol:
@iamareebjamal what do you say about [PhilJay/MPAndroidChart](https://github.com/PhilJay/MPAndroidChart)? shall I send PR to your repo? Or will you implement as most of the functions are similar? So it will be easier for you.

iamareebjamal:
I'll do it :+1:

iamareebjamal:
device-2017-07-21-034359

@mejariamol Please review now

mariobehling:
Please resolve conflicts.

PR #291, opened 2017-07-10
feat: Show formatted date of purchased ticket

### LINKED ISSUE #289: Show formatted Date in Attendee Check In Bottom Sheet
Currently, ISO format date is shown in the Attendee Check-In Bottom Sheet

We need to show the formatted date

iamareebjamal:
Fixes #289 

- Add formatted date in Attendee Check-In layout

device-2017-07-11-035541

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/app/utils/DateUtils.java
+++ b/app/src/main/java/org/fossasia/openevent/app/utils/DateUtils.java
@@ -15,6 +15,7 @@ public class DateUtils {
     public static final String FORMAT_12H = "hh:mm a";
     public static final String FORMAT_24H = "HH:mm";
     public static final String FORMAT_DATE_COMPLETE = "EE, dd MMM yyyy";
+    public static final String FORMAT_DAY_COMPLETE = "HH:mm, EE, dd MMM yyyy";
     public static final String FORMAT_DATE = "d MMM";
 
     private static boolean showLocalTimeZone;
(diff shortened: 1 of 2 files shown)


niranjan94:
@iamareebjamal what timezone is the date/time displayed in ?

mejariamol:
@niranjan94 Yeah. I had some doubts about the same.
1. As app will be used by an organizer so event's timezone shall be used in the whole app for time display right?
2. When an attendee books a ticket for an event, in which timezone the date is stored for order completion?

niranjan94:
1. Yes. Use event timezone throughout. But ensure the user (organizer)
knows what timezone are dates and times displayed in. Give an option for
the organizer to change display timezone.
2. The timezone information is part of the api date time response. Use
that.

iamareebjamal:
@niranjan94 The timezone is event timezone throughout the app. And the current code supports local date time conversion, just not the UI component to do so. Should we just add option to switch local and event timezone or should we add option to choose any timezone? I think we should let local timezone from Android decide this behaviour as user can change the timezone from there?

And the problem in showing the timezone part is that if we include the timezone format in Date Format in Java, it automatically converts it to local time. There is no clean way to do it using the API date response, perhaps if we add a different function just to extract timezone information from the date?

iamareebjamal:
@mejariamol For 2nd point, it's handled by the server

iamareebjamal:
@mejariamol @arp95 @Shailesh351 @shreyasnbhat @rohanagarwal94 Please review

mejariamol:
If styling can be taken in the PR, I think texts would look better if non all caps lock. For examle
Completed, Jul etc.. What do you think?

iamareebjamal:
I like to think it this way, that they are representing one of the constant states, paid, free, donation, completed, etc. So I think we should use caps, but if not, we can discuss on another issue as this is date related

PR #295, opened 2017-07-11
feat: Simplify data flow of event data

### LINKED ISSUE #268: Simplify MainActivity
Current implementation is too complex and data flow is not unidirectional

- Make data flow simpler throughout components
- Remove complexity from views, especially MainActivity

iamareebjamal:
Fixes #268 

- Implement typesafe Bus
- Replace listeners with bus implementation
- Refactor code to simplify data flow


app/src/main/java/org/fossasia/openevent/app/event/detail/EventDetailFragment.java
```diff
@@ -145,7 +130,6 @@ public void onRefreshComplete() {
 
     @Override
     public void showEvent(Event event) {
-        listener.onEventLoaded(event, false);
```
mejariamol: this was responsible for setting drawer header values (event name, time) when the app is restarted with stored event id. (event key in prefs not equal to -1).
please add this functionality accordingly

iamareebjamal: Yeah, I wanted to discuss this. Shouldn't main presenter load the event independently from detail presenter?

mejariamol: yeah. it can.
but I am thinking, why the main presenter should load the event if it can use info from the detail fragment which will be loaded initially for sure. ??

iamareebjamal: There is no guarantee that when this implementation will change. It is not really great to depend on a fragment to trigger the event being loaded for the main presenter, because, suppose if we remove dashboard as first page or remove the event loading logic from t completely (maybe because analytics API has been added), we'll need to implement it anyway. IMO, this should be independent

mejariamol: yeah. the main presenter should load required details separately then.

iamareebjamal: Doing the changes

iamareebjamal:
@mejariamol Please review

PR #312, opened 2017-07-17
feat: Use JWT to match existing and authenticated user

### LINKED ISSUE #305: Parse JWT token to get user identity
- Add parsing logic
- Add tests for parsing logic
- Remove user fetch request and use JWT id instead

I'm on it

iamareebjamal:
Fixes #305

- Extract identifier from JWT
- Add test for identifier extraction
- Use identifier to discern the user is same or not
- Fix tests

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/app/data/LoginModel.java
+++ b/app/src/main/java/org/fossasia/openevent/app/data/LoginModel.java
@@ -8,7 +8,6 @@ import org.fossasia.openevent.app.data.db.contract.IDatabaseRepository;
 import org.fossasia.openevent.app.data.models.Login;
 import org.fossasia.openevent.app.data.models.User;
 import org.fossasia.openevent.app.data.network.EventService;
-import org.fossasia.openevent.app.data.repository.contract.IEventRepository;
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/fossasia/openevent/app/data/LoginModel.java
```diff
@@ -48,10 +45,11 @@ public Completable login(String username, String password) {
         return eventService
             .login(new Login(username, password))
             .flatMapSingle(loginResponse -> {
-                utilModel.saveToken(loginResponse.getAccessToken());
+                String token = loginResponse.getAccessToken();
+                utilModel.saveToken(token);
                 utilModel.addStringSetElement(Constants.SHARED_PREFS_SAVED_EMAIL, username);
 
-                return isPreviousUser();
+                return isPreviousUser(token);
```
mejariamol: what happens when token is null?

iamareebjamal: Never can be

iamareebjamal: RxJava never transmits null value, onError is thrown instead

PR #313, opened 2017-07-17
fix: Extract common string from loop and add search tests

### LINKED ISSUE #310: Fix repeated queries in SearchUtils and add tests
- SearchUtils contains code which gets executed over a loop of keys. To trim the query in each loop is wasteful of resources
- Add tests to verify behaviour of this class

iamareebjamal:
Fixes #310 

- Extract common string from search loop and add tests


iamareebjamal:
@mejariamol Yeah it got missed while testing. Adding it

PR #314, opened 2017-07-17
feat: Add sentry integration

### LINKED ISSUE #302: Add sentry support in the app
https://docs.sentry.io/clients/java/modules/android/

- Create a Timber tree for logging
- Create a proxy for each exception to go though

I'm working on it

iamareebjamal:
Fixes #302 

- Add sentry integration
- Add contextmanager to set and destroy sentry context
- Redirect error and info logs to sentry

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,7 @@ Our chat channel is on gitter here: https://gitter.im/fossasia/open-event-orga-a
 - [DBFlow](https://github.com/Raizlabs/DBFlow)
 - [FastAdapter](https://github.com/mikepenz/FastAdapter)
 - [Leakcanary](https://github.com/square/leakcanary)
+- [Sentry](https://github.com/getsentry/sentry-java)
 - [Timber](https://github.com/JakeWharton/timber)
 - Testing:
   - [JUnit4](https://github.com/junit-team/junit4)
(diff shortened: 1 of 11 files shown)


iamareebjamal:
@niranjan94 The sentry cli (in order to upload the mappings) require a sentry.properties containing project info and auth token. But we can't include it in the repo itself. However, the release build will fail without it. Any suggestions?

niranjan94:
@iamareebjamal we can have it as a travis env variable and create the `sentry.properties` file with it in travis.yml.

niranjan94:
Though, note that the env variable will be unavailable for PRs.

iamareebjamal:
I think it will be better if we make the uploading option configurable, only running if user wants it?

niranjan94:
Why do we want to upload PRs ? And we don't need release builds for PRs. 
Also, make sentry optional. Build shouldn't fail because sentry key doesn't exist. Nor should the app crash.

iamareebjamal:
I didn't mean to say that we want to upload mappings on PR. I just meant to disable the option to upload mappings until the user explicitly enables it. 

About making sentry optional, should I create a new build variant for it?

iamareebjamal:
@niranjan94 I have disabled the auto upload feature by default. The app does not crash if sentry key isn't present. The library is made that way. 

The build was only failing because the cli was trying to upload the mappings to sentry server, which is required for proguard based projects, but now user will just have to change that attrbute in build.gradle if he wants to enable it

PR #316, opened 2017-07-18
chore: single base interface for presenters

### LINKED ISSUE #315: Making single base interface for all presenters
Currently, there are two base interfaces for presenter - IPresenter and IDetailPresenter. 

Expected: Make single base interface for presenter creating separate method for attaching key to the detail presenter. So that single base interface can be kept for presenters.

Sending a PR

mejariamol:
fixes #315

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/app/event/detail/EventDetailFragment.java
+++ b/app/src/main/java/org/fossasia/openevent/app/event/detail/EventDetailFragment.java
@@ -81,7 +81,8 @@ public class EventDetailFragment extends BaseFragment implements IEventDetailVie
         if (arguments != null)
             initialEventId = arguments.getLong(EVENT_ID);
 
-        eventDetailPresenter.attach(initialEventId, this);
+        eventDetailPresenter.attach(this);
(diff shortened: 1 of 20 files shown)


mejariamol:
@iamareebjamal please review

app/src/main/java/org/fossasia/openevent/app/event/detail/EventDetailFragment.java
```diff
@@ -81,7 +81,8 @@ public void onCreate(Bundle savedInstanceState) {
         if (arguments != null)
             initialEventId = arguments.getLong(EVENT_ID);
 
-        eventDetailPresenter.attach(initialEventId, this);
+        eventDetailPresenter.attach(this);
+        eventDetailPresenter.attachKey(initialEventId);
```
iamareebjamal: This has to repeat in each class, which makes it automatically an anti-pattern. Clearly, this always needs to be called together and thus remain in a single method

mejariamol:
Not required for now. Could implement presenter loader without this. Hence closing it.

PR #317, opened 2017-07-18
feat: presenters via loaders

### LINKED ISSUE #196: Separate initial data load part from ui interation in presenters

mejariamol:
fixes #196

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/app/common/BaseActivity.java
+++ b/app/src/main/java/org/fossasia/openevent/app/common/BaseActivity.java
@@ -1,0 +1,58 @@
+package org.fossasia.openevent.app.common;
+
+import android.os.Bundle;
+import android.support.annotation.Nullable;
+import android.support.v4.app.LoaderManager;
+import android.support.v4.content.Loader;
+import android.support.v7.app.AppCompatActivity;
+
+import org.fossasia.openevent.app.common.contract.presenter.IBasePresenter;
+import org.fossasia.openevent.app.common.loader.PresenterLoader;
+
+import dagger.Lazy;
+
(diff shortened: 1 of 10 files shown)


mejariamol:
@iamareebjamal please review

mejariamol:
@iamareebjamal please test.

iamareebjamal:
@mejariamol I did, and still the resources are loaded again from db. But I guess this is not fault of the loader, it is presenter which will nevertheless load from db when start is called. Please create another issue for that.

About this PR, everything is perfect except that BaseActivity and BasePresenter have a lot of functionality in common. I think it'll be better if they share a common base class

mejariamol:
>But I guess this is not fault of the loader, it is presenter which will nevertheless load from db when start is called. Please create another issue for that.

yeah. Will create an issue for that once this is merged. :+1:

iamareebjamal:
OK, so just move common functions from base fragment and activity in another class

mejariamol:
>OK, so just move common functions from base fragment and activity in another class

Actually, this can't be more abstracted. As both the classes need to extend different classes - Fragment and AppCompatActivity. So can't make another base class which will contain common function and extend these classes by that class because a class cannot extend more than one class.

Any suggestions? @iamareebjamal

iamareebjamal:
True, my mistake

PR #320, opened 2017-07-19
feat: Create a LoaderHandler for lifecycle management

### LINKED ISSUE #319: Move common loader handling code in a composition
Create a lifecycle handler which handles common loader management for both activity and fragment

iamareebjamal:
Fixes #319

- Create lifecycle package
- Create common PresenterProvider interface
- Move common loader handling code from activity/fragment
- Apply annotations to check calling super
- Update views with presenter getter


iamareebjamal:
@mejariamol Please review

PR #321, opened 2017-07-19
feat: Add validation by data binding in login layout

### LINKED ISSUE #265: Use DataBinding in login layout
- Use data binding in login layout
- Enable validation for email and password with appropriate error message handling

iamareebjamal:
Fixes #265 

- Add data binding to login layout
- Rename layouts for better binding name
- Add data binding validation library
- Validate email, password length and base url


app/src/main/res/layout/login_activity.xml
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0" encoding="utf-8"?>
```
mejariamol: actually, we are following naming style `fragment_*` , `activity_*` and `content_*` for layouts

app/src/main/java/org/fossasia/openevent/app/login/LoginActivity.java
```diff
@@ -102,11 +84,13 @@ protected void onStart() {
             }
         });
 
-        btnLogin.setOnClickListener(v -> {
+        binding.loginContent.btnLogin.setOnClickListener(view -> {
+            if (!validator.validate())
```
mejariamol: this return true for `https://....com/api/v1/` which we are using currently. Although we can just use `https://....com` as a base URL because HostSelectionInterceptor takes care of the suffix section using original url passed to retrofit from `Constants.BASE_URL`.

app/src/main/res/layout/login_activity.xml
iamareebjamal: This makes it better when generating layout bindings. When using layout binding, it won't matter what is the naming style. Furthermore, if we follow the project style of package by feature. login_* should be followed since activity_* is package by layer

app/src/main/java/org/fossasia/openevent/app/login/LoginActivity.java
iamareebjamal: You mean it is showing error?

mejariamol: Could not able to log in with `https://....com/api/v1/` since validator doesn't validate this pattern. You may just change the default base URLs to `https://....com` or validate `https://....com/api/v1/`

iamareebjamal: It was due to the hidden field. I have removed the validation for now. Will move the handling logic in presenter in another PR and then add it

mariobehling:
Please resolve merge conflicts.

PR #323, opened 2017-07-19
feat: Make StickyHeaderAdapter generic

### LINKED ISSUE #322: Make StickyHeaderAdapter generic
- Remove instance checks in StickyHeaderAdapter
- Make it generic for future common use

iamareebjamal:
- Create header provider interface
- Remove instance checks from adapter
- Add generic type

Fixes #322


app/src/main/java/org/fossasia/openevent/app/data/models/Attendee.java
```diff
@@ -93,34 +94,7 @@ public Attendee withCheckedIn() {
...
-                    )
-            ) == 0 ?
-                email
-                    .toLowerCase()
-                    .compareTo(
-                        other
-                            .getEmail()
-                            .toLowerCase()
-                    ) : compareLastName : compareFirstName;
+        return getFirstName().compareTo(other.getFirstName());
```
mejariamol: Couldn't get this part? lowercase is also removed??

iamareebjamal: Will add lowercase but when is first name ever in lower case

But why did we need email and last name comparison?

mejariamol: >but when is first name ever in lower case

that is why these must be converted to lowercase before comparison.

>But why did we need email and last name comparison?

when firstnames are same, then last names will be compared and similarly emails

iamareebjamal: Yeah I know, but do we really need that strict sorting? 

About the lowercase. The first name is never in lower case, neither is the last name.

mejariamol: >The first name is never in lower case, neither is the last name.

before comparing, the keys should be converted to the same case right? As `Bikash` will come before `amol` if `toLowerCase` is not used while comparing.

iamareebjamal: The point I'm trying to convey is that is there any case where the first name will be `amol`. It will always be `Amol`

mejariamol: screenshot_2017-07-20-13-10-52

with only firstname comparison `Amol Mejari` would appear on top of `amol lowercase`. Please keep the complete comparison as it is just a few lines of code, and not gonna run all the cases each time.

iamareebjamal: OK

iamareebjamal:
@mejariamol Please review now

app/src/main/java/org/fossasia/openevent/app/utils/CompareUtils.java
```diff
@@ -0,0 +1,24 @@
...
+                int current = mapper.apply(one).toLowerCase().compareTo(mapper.apply(two).toLowerCase());
+                if (current != 0)
+                    return current;
+            } catch (Exception e) {
+                // No-Op Can't happen
+                Timber.wtf(e);
+            }
+        }
+
+        return 0;
```
mejariamol: should not return 0 as it shall make two different attendees equal if the names are matched. As this is returned in general `compareTo` interface.

iamareebjamal: This will return 0 iff name, lastname and email of attendee are same

mejariamol: yeah. but we don't want that. As it is in `compareTo` interface. This shall give equality for two different attendees if firstname, lastname and emails are matched in lowercase. For equality we should consider attendee's id only right? so here only either 1 or -1 should be returned.

iamareebjamal: That is `equals` method you are talking about. This is strictly used for sorting

PR #333, opened 2017-07-20
fix: removes unnecessary event list page load on start

### LINKED ISSUE #332: Unnecessary event list page load on MainActivity start
Currently MainActivity loads event list page on start irrespective of event ID stored in prefs.

Expected:
On start, presenter should check stored event id and accordingly should load correct fragment.

Sending a PR

mejariamol:
fixes #332


mejariamol:
@iamareebjamal please review

PR #334, opened 2017-07-20
[Later] fix: avoid detaching presenter by loader

### LINKED ISSUE #326: Verify correct detach call on presenter
Loaders automatically remove the instance of item when the loader container is actually getting destroyed. Currently, several detach calls are made in fragment and activities stopping any ongoing task, which defeats the purpose of retaining presenters.

- View should get detached for each fragment/activity
- The containing composite disposable should only be disposed when the view is actually destroyed

mejariamol:
fixes #326

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/app/common/lifecycle/loader/PresenterLoader.java
+++ b/app/src/main/java/org/fossasia/openevent/app/common/lifecycle/loader/PresenterLoader.java
@@ -23,7 +23,6 @@ public class PresenterLoader<T extends IBasePresenter> extends Loader<T> {
     @Override
     protected void onReset() {
         super.onReset();
-        presenter.detach();
         presenter = null;
     }
 


mejariamol:
@iamareebjamal please review

iamareebjamal:
Actually, this issue is much deeper. See the description about destroying of view

mejariamol:
>View should get detached for each fragment/activity
  The containing composite disposable should only be disposed when the view is actually destroyed

both of them are carried out under presenter's detach method. And presenter's detach is called in onStop of fragment and activity. Didn't get what you are saying.. Please elaborate. @iamareebjamal

iamareebjamal:
There are 2 things -

Some things have to be stopped for even the rotations, and other things only when the view is actually being destroyed. You have removed the loader's detach method which gets called when the view is actually destroyed

mejariamol:
@iamareebjamal still not clear..

iamareebjamal:
Will write a detailed explanation in a few hours :smile:

mejariamol:
@iamareebjamal waiting for your reply. :smile:

iamareebjamal:
OK, so consider this:

There are 2 cases where onStop is called:

1. Destructive : View is destroyed. No data is needed to be saved. When anyone presses back button or switches fragment. Detach should be called and data should not be saved
2. Disruptive : Rotation or any other pausing of fragment. Detached should be called because a new fragment will nevertheless be created and it must be attached to the presenter via attach(id, view) BUT the data must be saved and replayed

mejariamol:
@iamareebjamal So you mean,

for case 1: null the view and dispose the CompositeDisposable
for case 2: null the view but don't dispose the CompositeDisposable  ??

iamareebjamal:
Yes

mejariamol:
closing this for now. The issue is little deeper.

PR #344, opened 2017-07-22
feat: Add settings fragment and option to show local dates

### LINKED ISSUE #292: Create Settings Page and add timezone changing settings
- Create Settings Page
- Add option to switch from event timezone to local timezone

iamareebjamal:
Fixes #292

- Add RxPreferences
- Add Settings Fragment
- Add option to pick if local dates are to be shown

photo6316454234832349109


iamareebjamal:
Picking local timezone will show INVALID date as order time is not ISO861 compliant

iamareebjamal:
@mejariamol Please review

mejariamol:
@iamareebjamal please add screenshot

app/src/main/java/org/fossasia/openevent/app/main/MainActivity.java
```diff
@@ -114,7 +115,7 @@ public void onDrawerClosed(View drawerView) {
 
                 if (id == R.id.nav_logout)
                     showLogoutDialog();
-                else if (eventId != -1)
+                else if (eventId != -1 || id == R.id.nav_settings)
```
mejariamol: this case is not required now, as we are hiding event specific options already if event id is not stored.

iamareebjamal: But this is not an event specific item

mejariamol: yes, so?

iamareebjamal: So we should also show it if the event id is not set

mejariamol: yes, you have added it to the `mainMenu` group so it will be always shown. Only visibility condition is applied to the `subMenu` which contains event specific items.

iamareebjamal: But the loadFragment won't be called if I remove the or condition if eventId == -1 which we do not want

mejariamol: remove the complete condition. just `else` is enough there.

iamareebjamal: OK

iamareebjamal:
photo6316454234832349109

PR #350, opened 2017-07-23
Event Detail Page UI Modification

### LINKED ISSUE #349: Event Detail Page Modification
By now, a good analytics is shown on Event detail page but the items are little less organized. 

The modifications can be:
1. Total sale star is overshown compared to the info it shows. Change it.
2. Move sales chart below the total sale
3. Square ticket cards.
4. uniform sized sub headings

@iamareebjamal please suggest if I missed anything.

taking this up.

eventbrite screenshot:

dashboard

mejariamol:
fixes #349 

screenshot_2017-07-24-12-07-21

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/app/event/detail/EventDetailPresenter.java
+++ b/app/src/main/java/org/fossasia/openevent/app/event/detail/EventDetailPresenter.java
@@ -55,7 +55,10 @@ public class EventDetailPresenter extends BaseDetailPresenter<Long, IEventDetail
 
         chartAnalyser.loadData(getId())
             .compose(disposeCompletable(getDisposable()))
-            .subscribe(() -> chartAnalyser.showChart(getView().getChartView()), Logger::logError);
+            .subscribe(() -> {
+                getView().showChart(true);
+                chartAnalyser.showChart(getView().getChartView());
+            }, throwable -> getView().showChart(false));
 
         eventRepository
             .getEvent(getId(), forceReload)
--- a/app/src/main/java/org/fossasia/openevent/app/event/detail/EventDetailFragment.java
+++ b/app/src/main/java/org/fossasia/openevent/app/event/detail/EventDetailFragment.java
@@ -162,4 +162,9 @@ public class EventDetailFragment extends BaseFragment<IEventDetailPresenter> imp
     public LineChart getChartView() {
         return binding.ticketAnalytics.chart;
     }
+
+    @Override
+    public void showChart(boolean show) {
+        ViewUtils.showView(binding.ticketAnalytics.chartBox, show);
+    }
 }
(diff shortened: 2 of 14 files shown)


mejariamol:
@iamareebjamal please review

iamareebjamal:
I think using cards is better

mejariamol:
I tried cards, but then there become so many cards and it doesn't look uniform. Cards are good for the same type of items. Here we have different items. So I think cards won't look good.

mejariamol:
@iamareebjamal how about this?

img_20170724_002118_466

mejariamol:
@iamareebjamal please review. 
Chart visibility condition applied.

iamareebjamal:
The card shadow will look good on the layout. Don't use the rounded corner. Your layout is looking good, the card shadow will make it better. Also, Material Design guidelines say that cards are best for different item containers

 Also, don't make the toolbar and top surface same color. Perhaps make the toolbar completely white and the top surface a little more whiter. This is looking a tad bit greyer.

mejariamol:
@iamareebjamal how is this?

screenshot_2017-07-24-07-43-56

iamareebjamal:
Looks great. Just ensure that the toolbar color and surface color of cards is different :+1:

mejariamol:
>Just ensure that the toolbar color and surface color of cards is different

toolbar color and top surface color should be same. bottom color will be different.

>The card surface is a very grey. #eee should be fine for it

actually, it is #f2f2f2 same as toolbar. :smile:  thats why I told you to test it on device.

mejariamol:
And card view is not used as fitting to the width is better. bottom shadow is sufficient like toolbar.

iamareebjamal:
CardView can also fit to width
device-2017-04-07-145211
device-2017-04-07-145419

How the bottom shadow is generated? 

Also, Toolbar should be discernable from other views. Either make the toolbar whiter or the panels. See the screenshots above

mejariamol:
>How the bottom shadow is generated?

two frame layouts are used in a linear layout.
will use card view instead. :+1:

iamareebjamal:
@mejariamol Looks good. Please create a new comment when mentioning. I was not aware that this was changed

app/src/main/java/org/fossasia/openevent/app/event/detail/EventDetailPresenter.java
```diff
@@ -55,7 +55,13 @@ public void loadDetails(boolean forceReload) {
 
         chartAnalyser.loadData(getId())
             .compose(disposeCompletable(getDisposable()))
-            .subscribe(() -> chartAnalyser.showChart(getView().getChartView()), Logger::logError);
+            .subscribe(() -> {
+                getView().showChart(true);
+                chartAnalyser.showChart(getView().getChartView());
+            }, (throwable) -> {
+                Logger.logError(throwable);
```
iamareebjamal: Throwable can be put out of the brackets. And Logger is only used when we don't provide any action on error. It's just a proxy

PR #360, opened 2017-07-25
feat: Add delete ticket functionality

### LINKED ISSUE #359: Add delete ticket option
Deleting a ticket is much more basic than creating one, so add this functionality first

iamareebjamal:
- Remove automatic attendee download on getting tickets
- Add delete method in event service, ticket repository and presenter
- Use data binding to show intermediate state of ticket deleting

Fixes #359

screenshot-1500955963775

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/app/common/contract/Pipe.java
+++ b/app/src/main/java/org/fossasia/openevent/app/common/contract/Pipe.java
@@ -1,0 +1,7 @@
+package org.fossasia.openevent.app.common.contract;
+
+public interface Pipe<T> {
+
+    void push(T item);
+
+}
--- a/app/src/main/java/org/fossasia/openevent/app/common/data/models/Ticket.java
+++ b/app/src/main/java/org/fossasia/openevent/app/common/data/models/Ticket.java
@@ -46,6 +46,8 @@ public class Ticket implements Comparable<Ticket> {
     public String salesEndsAt;
     public Integer minOrder;
 
+    public final ObservableBoolean deleting = new ObservableBoolean();
+
     @Relationship("event")
     @ForeignKey(stubbedRelationship = true, onDelete = ForeignKeyAction.CASCADE)
     public Event event;
--- a/app/src/main/java/org/fossasia/openevent/app/common/data/network/EventService.java
+++ b/app/src/main/java/org/fossasia/openevent/app/common/data/network/EventService.java
@@ -9,8 +9,10 @@ import org.fossasia.openevent.app.common.data.models.dto.LoginResponse;
 
 import java.util.List;
 
+import io.reactivex.Completable;
 import io.reactivex.Observable;
 import retrofit2.http.Body;
+import retrofit2.http.DELETE;
 import retrofit2.http.GET;
 import retrofit2.http.PATCH;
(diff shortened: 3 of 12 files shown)


mejariamol:
@iamareebjamal
what if the ticket is booked by the attendees?

iamareebjamal:
That should be handled by the server once the order API is complete. For the app, the constraints make the attendee get deleted as well, but as soon as the event detail or attendee page is opened, it is reloaded from the server. 

In short, the server can have detached attendees as of now, but we must implement basic client CRUD

mejariamol:
@iamareebjamal let's take Niranjan's view on this

@niranjan94 do we want this feature in the app? Ticket deletion.

niranjan94:
I don't think ticket deletion is to be done at this point. Because, this still needs a lot to think through in terms of business logic. Some of the main things are:
- Handling/triggering refunds
- Cancellation charges (and charges based on when they're cancelling)
- What if it was paid in cash at the venue ? How do we handle refunds.
- Does deleted tickets open up instantly for new booking or is there a wait period

niranjan94:
So, have the app code if you want. But hide the delete button for now.

mejariamol:
>Does deleted tickets open up instantly for new booking or is there a wait period

@iamareebjamal @niranjan94 Some confusion there.
I think this feature is about deleting the ticket type of the event. right? @iamareebjamal 
@niranjan94 might have misunderstood it with individual attendee deletion.

niranjan94:
@mejariamol updated my comment. Yes. I had mis-understood. But my original concerns still remain.

app/src/main/java/org/fossasia/openevent/app/common/contract/Pipe.java
```diff
@@ -0,0 +1,7 @@
+package org.fossasia.openevent.app.common.contract;
+
+public interface Pipe<T> {
```
mejariamol: where is this used?

iamareebjamal: To push the ticket from viewholder to adapter. It is essentially a Java 8 or Rx consumer, but Java 8 one does not work on Android and Rx one needs to throw exception.

iamareebjamal:
@mejariamol Please review this

app/src/main/java/org/fossasia/openevent/app/common/data/models/Ticket.java
```diff
@@ -42,6 +43,8 @@
     public int minOrder;
     public String salesStartsAt;
 
+    public final ObservableBoolean deleting = new ObservableBoolean();
```
mejariamol: shouldn't it be `isDeleting`?

iamareebjamal: boolean fields should be deleting and getters/setters be isDeleting setDeleting

iamareebjamal:
@niranjan94 Please merge

mariobehling:
Please resolve conflicts.

niranjan94:
@iamareebjamal please resolve conflicts here :)

iamareebjamal:
@niranjan94 Done :smile:

PR #369, opened 2017-07-26
fix: event bus behaviour

### LINKED ISSUE #368: Repair event bus behaviour
Currently, BehaviourSubject is used in Bus implementation. But it is causing issue:
On new user login, it emits the last event emitted and causes to open event dashboard of wrong event.

Expected: Replace it with PublishSubject again. 

Sending a PR

mejariamol:
- used `PublishSubject` instead of `BehaviourSubject` in bus

fixes #368

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/app/common/data/Bus.java
+++ b/app/src/main/java/org/fossasia/openevent/app/common/data/Bus.java
@@ -6,10 +6,10 @@ import org.fossasia.openevent.app.common.data.models.Event;
 import javax.inject.Inject;
 
 import io.reactivex.Observable;
-import io.reactivex.subjects.BehaviorSubject;
+import io.reactivex.subjects.PublishSubject;
 
 public class Bus implements IBus {
-    private static BehaviorSubject<Event> eventIdPublisher = BehaviorSubject.create();
+    private static PublishSubject<Event> eventIdPublisher = PublishSubject.create();
 
     @Inject
     public Bus() {}


mejariamol:
@iamareebjamal please review

iamareebjamal:
BehaviourSubject is needed. 

The solution can be that you emit -1 when logging out 

If we use PublishSubject we will not get selected value in many places because the subscription did not happen before emitting the value,

mejariamol:
>If we use PublishSubject we will not get selected value in many places because the subscription did not happen before emitting the value,

Where do we need this behaviour now in the app?

iamareebjamal:
Anywhere. We should get the previously pushed event. This is used on ticket creation page now

mejariamol:
but we are storing it in the prefs right?

iamareebjamal:
We can't store event in prefs

mejariamol:
I mean we are storing the event id.

iamareebjamal:
So we need to load event from db everytime someone changes it?

mejariamol:
yeah. I am not getting your point. does the behavior subject is able to emit the last emitted value always. I mean after the app is closed for few hours.

iamareebjamal:
Suppose I want what is the current selected event. Why should I load it from the sharedpreference and memory when it is already pushed in the bus. Obviously when the app is opened for the first time, it will be loaded from db, but from then on, bus will be used to get it. If you use PublishSubject instead and I subscribe later, it'll never get the last emitted value

mejariamol:
yeah. but when I tested earlier behavior subject emits the last item as it is subscribed. this causes two times dashboard load when storedEventId is not -1. I will check again.

iamareebjamal:
Nope, that was another issue fixes in #372

mejariamol:
will be handled in #376 and #375 . Closing this

PR #371, opened 2017-07-26
chore: removes unnecessary methods from presenters

### LINKED ISSUE #363: Remove unnecessary methods from presenters
Presenters contain unnecessary methods of attach and detach without doing anything except calling super method. They should be removed

mejariamol:
- removes attach and detach methods from presenters with the only super method.

fixes #363


mejariamol:
@iamareebjamal please review

PR #372, opened 2017-07-27
feat: Enable lazy initialisation of app

### LINKED ISSUE #367: Lazy initialise items while application creation
Various items are initialised onCreate of application, making it slow. Items should be lazily initialized instead for gaining startup performance

I'm working on it

iamareebjamal:
- Move sentry config in release block
- Make CurrencyUtils and ContextManager reactive
- Fix multiple loading of dashboard
- Fix proguard rules for release build

Fixes #367

### DIFF
--- a/app/proguard-rules.pro
+++ b/app/proguard-rules.pro
@@ -31,7 +31,7 @@
 
 -dontwarn javax.annotation**
 
--keepclassmembers class org.fossasia.openevent.app.data.models.** {
+-keep class org.fossasia.openevent.app.common.data.models.** {
   *;
 }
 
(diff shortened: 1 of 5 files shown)


mejariamol:
@iamareebjamal 
Still, it loads dashboard two times on rotation. Please take a look.

iamareebjamal:
The problem is deeper than that. 

1. The EventFragment itself loads the presenter without needing the main presenter to push the changes on rotation (because of the presenter persisting via loader)
2. The click handling is also done by the bus

Thus, event presenter is indeed started 3 times. There must be a solution for this but in another issue.

iamareebjamal:
@mejariamol Let's discuss here - #376 

BehaviourSubject indeed loads 3 times, but PublishSubject would do it twice which is still not acceptable.

iamareebjamal:
@mejariamol It does remove multiple load bug. The rotation bug is because of lifecycle and bus. On starting, loadFragment was unanimously called twice if eventId was not -1, this fixes it

Issue #376, opened 2017-07-27
Manage event selection of app


iamareebjamal:
Bus handles click on the selection of event and the event fragment itself triggers the loading of event dashboard. Discuss and plan solution for this

mejariamol:
@iamareebjamal 
Currently, the bus is used for 2 purposes:
1. pushing event from events adapter to the presenter to open dashboard
2. handling the state of the main activity using `storedEventId`.

I think 2nd should be handled in the activity itself without the use of the bus. this will solve #375 too.

iamareebjamal:
I think 1st should be handled by something else. 

Because we do want a stored session event which is reactive. What solution do you propose we can use to make it happen without bus? The problem is with the clicks opening the fragment through bus.

Bus should only update the header view and currency and nothing else. The trigger on event fragment should be handled by something which is not connected to bus

mejariamol:
>The problem is with the clicks opening the fragment through bus.

what problem?

iamareebjamal:
The repeated loading of event dashboard

mejariamol:
No. problem is in the presenter's `onStart` which is handling state of the main activity.

mejariamol:
Initial state of the main activity should be handled in onCreate of main activity

mejariamol:
using the nullity of savedInstanceState

iamareebjamal:
The thing is, bus is also transmitting a value. Which is valid for setting header info, but not for the loading of fragment. The fragment itself is loaded on rotation, so if the bus is not used to open the dashboard, it'll work fine. Yes, your solution is obviously needed, but it is related to #325 

Try by adding take(1) to bus, it will only use 1st transmission from bus and after that rotation will load the fragment automatically

PR #364, opened 2017-07-26
feat: Add create ticket feature

iamareebjamal:
Fixes #267 

- Add ticket POST methods in service and repository
- Add ticket database change listener
- Create view and presenter for creating tickets
- Add ticket binder for binding layout to ticket fields
- Add create ticket FAB in tickets page
- Fix format for ISO date in DateUtils
- Add base dialog fragment
- Add appropriate icons
- Fix event serialization by ignoring it
- Make ticket fields class type in order to ignore nulls
- Create intermediate field for observing creation

screenshot-1501061872745
screenshot-1501061881288

Issue #353, opened 2017-07-24
Show Ticket Details in Tickets Page


iamareebjamal:
When clicked on a ticket, ticket details should be shown in a BottomSheetDialog

PR #377, opened 2017-07-27
fix: retain main activity state on rotation

### LINKED ISSUE #375: Handling rotation for Main Activity
To be done after #372 

Currently, on rotation main activity opens event dashboard.

Expected:
If any fragment is opened already, it should remain on rotation too.

Taking this up.

mejariamol:
- used publish subject instead of behavior subject for bus
- shifted initial page load logic on main activity start to main activity's onCreate
- bus is not used to show header title any more

fixes #375

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/app/common/data/Bus.java
+++ b/app/src/main/java/org/fossasia/openevent/app/common/data/Bus.java
@@ -6,10 +6,10 @@ import org.fossasia.openevent.app.common.data.models.Event;
 import javax.inject.Inject;
 
 import io.reactivex.Observable;
-import io.reactivex.subjects.BehaviorSubject;
+import io.reactivex.subjects.PublishSubject;
 
 public class Bus implements IBus {
-    private static BehaviorSubject<Event> eventIdPublisher = BehaviorSubject.create();
+    private static PublishSubject<Event> eventIdPublisher = PublishSubject.create();
 
     @Inject
     public Bus() {}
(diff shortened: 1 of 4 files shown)


mejariamol:
@iamareebjamal 
have replaced behavior subject to publish subject. please test on a device. we may restore it with behavior subject again with #376. please review.

iamareebjamal:
@mejariamol There's no point in changing something temporarily. What issue will it fix if it is PublishSubject?

mejariamol:
Behavior subject is causing issue for rotation by re-emitting the last item to load dashboard forcefully, right?

mejariamol:
I am just saying if we are using behavior subject then it should be used correctly.

iamareebjamal:
Yes, that's why we should fix the behavior rather than changing it to PublishSubject. If we want to use PublishSubject, then fine. But then we must implement a solution to the problem we are facing in #364 if BehaviorSubject is not used. It will break on merging on this PR

mejariamol:
> But then we must implement a solution to the problem we are facing in #364 if BehaviorSubject is not used

will have to use stored event id and get the event from the repository to set to ticket.

iamareebjamal:
Then what'd be the point of a bus?

mejariamol:
For event click on events page.

iamareebjamal:
That's not the use of bus. For that, we could even use action listeners. This is a workaround TBH, not a solution, so we should either complete the solution or else there'd be no point of the bus. I'll be thinking about the solution, meanwhile could you take up #353 

These bugs can be solved later, as the app is working fine without them. Probably when the feature set s stable, we can take these up.

iamareebjamal:
Handled in #389

PR #379, opened 2017-07-29
feat: Use ContentProvider trick to load context

### LINKED ISSUE #378: Use Firebase ContentProvider trick to load Context
https://firebase.googleblog.com/2016/12/how-does-firebase-initialize-on-android.html

This will reduce the dependency on context and make mocking easier

Working on it

iamareebjamal:
Fixes #378 

- Add OrgaProvider
- Change Context dependencies to preloaded static version


mariobehling:
Good job!

Please resolve conflicts.

PR #403, opened 2017-08-02
feat: Bind entities to layout during auth

### LINKED ISSUE #402: Bind entities during auth process
Use layout binding instead of manually setting properties

iamareebjamal:
- Rebase AuthModel to use model instead of strings
- Remove manual bindings from sign up and login presenter
- Make both presenters contain the model for binding
- Create binding with lambda for Go Action
- Fix AutoCompleteEditText next action
- Use 2 way binding to bind layout and entity
- Make login/signup button standard
- Use flat buttons instead of link text for switching
- Add animations while switching layouts
- Make AuthActivity style consistent with light theme
- Rename AuthActivity layout XML
- Make common base url layout
- Use binding to control visibility of url layout
- Remove redundant styles
- Refactor tests

Fixes #402


iamareebjamal:
screenshot-1501674976068
screenshot-1501675008314

iamareebjamal:
@mejariamol @Shailesh351 @shreyasnbhat Please review

mejariamol:
>Use flat buttons instead of link text for switching

Some suggestions for this change:

1. it doesn't seem to be a clickable view. 
2. use accent color (blue) as text color
3. All caps is not looking good.

iamareebjamal:
1. As shown in GIF, it is clickable
2. Flat buttons are to be of primary color and all caps

Using link text does not make sense here as it gives an impression that it is a link whereas it is not

app/src/main/java/org/fossasia/openevent/app/module/auth/login/LoginFragment.java
```diff
@@ -114,26 +89,11 @@ public void onStop() {
...
     }
 
-    private void setEditTextListener() {
-        binding.etBaseUrl.setOnEditorActionListener(this);
-        binding.emailDropdown.setOnEditorActionListener(this);
-        binding.etPassword.setOnEditorActionListener(this);
-    }
-
-    @Override
-    public boolean onEditorAction(TextView v, int actionId, KeyEvent event) {
```
iamareebjamal: Go button should only be added on the last field of the form, or else the next buttons won't work. And data binding should be used for it

mejariamol:
>As shown in GIF, it is clickable

yeah. but it must seem to be like a clickable. right? May be you can give some color to it.

iamareebjamal:
> yeah. but it must seem to be like a clickable. right? May be you can give some color to it.
http://www.materialdoc.com/flat-button/

mejariamol:
Okay. But I think at least giving accent color to the text color will be better. as currently, it doesn't appear to be clickable by its display.

iamareebjamal:
It is according to material specs, but will add it if you say so.

mejariamol:
>It is according to material specs, but will add it if you say so. 

please check if it looks better. If not, keep this one. may be with some extra top margin :+1:

iamareebjamal:
@mejariamol Please review now

PR #411, opened 2017-08-03
fix: Remove pagination from API requests

### LINKED ISSUE #409: Remove default pagination of 20 from API requests
As of now, not more than 20 resources can be loaded from the server due to default pagination. This is not the correct behavior for the application

Working on it

iamareebjamal:
Fixes #409

- Remove default pagination


iamareebjamal:
@mejariamol @Shailesh351 @shreyasnbhat Please review

PR #414, opened 2017-08-04
fix: disable background clicks when progress bar is shown

### LINKED ISSUE #412: Make background unclickable when progress bar is shown
Currently, all the views on the page are clickable while the progress bar is shown. 

Expected: When the progress bar is shown, make the page just viewable and disable clicks.

working on it.

mejariamol:
- modified progress bar views in all the layouts

fixes #412 

screenshot_2017-08-04-14-42-14

### DIFF
--- a/app/src/main/res/layout/activity_chart.xml
+++ b/app/src/main/res/layout/activity_chart.xml
@@ -20,19 +20,13 @@
         app:srcCompat="@drawable/ic_exit_full_screen"
         app:fabSize="normal" />
 
-    <LinearLayout
-        android:id="@+id/progressView"
+    <FrameLayout
+        android:id="@+id/progressBar"
         android:layout_width="match_parent"
         android:layout_height="match_parent"
-        android:background="@color/color_top_surface"
         android:visibility="gone">
-    <ProgressBar
-        android:id="@+id/progressBar"
-        style="?android:attr/progressBarStyle"
-        android:layout_width="match_parent"
-        android:layout_height="wrap_content"
-        android:layout_gravity="center" />
 
-    </LinearLayout>
+        <include layout="@layout/progressbar_layout" />
+    </FrameLayout>
 
 </android.support.design.widget.CoordinatorLayout>
--- a/app/src/main/res/layout/fragment_attendees.xml
+++ b/app/src/main/res/layout/fragment_attendees.xml
@@ -56,13 +56,14 @@
             <include layout="@layout/empty_layout" />
         </FrameLayout>
 
-        <ProgressBar
+        <FrameLayout
             android:id="@+id/progressBar"
-            style="?android:attr/progressBarStyle"
             android:layout_width="match_parent"
-            android:layout_height="wrap_content"
-            android:layout_gravity="center"
-            android:visibility="gone"
-            tools:visibility="visible" />
+            android:layout_height="match_parent"
+            android:visibility="gone">
+
+            <include layout="@layout/progressbar_layout" />
+        </FrameLayout>
+
     </android.support.design.widget.CoordinatorLayout>
 </layout>
--- a/app/src/main/res/layout/sign_up_fragment.xml
+++ b/app/src/main/res/layout/sign_up_fragment.xml
@@ -11,113 +11,117 @@
             type="org.fossasia.openevent.app.common.data.models.User" />
     </data>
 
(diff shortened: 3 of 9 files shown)


mejariamol:
@iamareebjamal please review.

app/src/main/java/org/fossasia/openevent/app/module/event/dashboard/EventDashboardFragment.java
```diff
@@ -143,7 +143,7 @@ private void setupRefreshListener() {
 
     @Override
     public void showProgress(boolean show) {
-        ViewUtils.showView(binding.progressBar, show);
+        ViewUtils.showView(binding.progressView, show);
```
iamareebjamal: Is it necessary to rename this? It is a progress bar after all. Also, we can create a common layout and just import it in each one

mejariamol:
@iamareebjamal please review

app/src/main/res/layout/sign_up_fragment.xml
```diff
@@ -11,113 +11,117 @@
             type="org.fossasia.openevent.app.common.data.models.User" />
     </data>
 
-    <android.support.v4.widget.NestedScrollView
+    <FrameLayout
```
iamareebjamal: Why is this wrapping needed?

mejariamol: Data binding only allows one data and one layout item

iamareebjamal: Wrapping in FrameLayout makes no change in data or layout count.

iamareebjamal: OK, I see what you are saying. But why wrap this in RelativeLayout? 

1. Use CoordinatorLayout as root
2. Do not use RelativeLayout. The layout can be directly imported without needing to be wrapped with any layout

mejariamol: >Do not use RelativeLayout. The layout can be directly imported without needing to be wrapped with any layout

directly using include was allowing directly binding it for setting visibility by setting id to it.

iamareebjamal: Still that is better because this method is wrapping it inside an unnecessary hierarchy, which is also bad for performance

mejariamol: @iamareebjamal please check my updated comment.

iamareebjamal: https://stackoverflow.com/questions/17653642/hide-included-layout-in-android

mejariamol: Using data binding it is not allowing? I will re check again. @iamareebjamal

mejariamol: @iamareebjamal using data binding to set the visibility of the included view directly is not working. Will have to go with this. Same is used for the empty layout too.

iamareebjamal: Then use CoordinatorLayout instead of FrameLayout

mejariamol:
@iamareebjamal please review now.

app/src/main/res/layout/activity_chart.xml
```diff
@@ -20,19 +20,12 @@
         app:srcCompat="@drawable/ic_exit_full_screen"
         app:fabSize="normal" />
 
-    <LinearLayout
-        android:id="@+id/progressView"
+    <RelativeLayout
```
iamareebjamal: This can be FrameLayout. No need of RelativeLayout here, RelativeLayout is expensive

app/src/main/res/layout/fragment_attendees.xml
```diff
@@ -56,13 +56,13 @@
...
             android:id="@+id/progressBar"
-            style="?android:attr/progressBarStyle"
             android:layout_width="match_parent"
-            android:layout_height="wrap_content"
-            android:layout_gravity="center"
-            android:visibility="gone"
-            tools:visibility="visible" />
+            android:layout_height="match_parent"
+            android:visibility="gone">
+        <include layout="@layout/progressbar_layout" />
```
iamareebjamal: Indentation is wrong

mejariamol:
@iamareebjamal please review.

PR #418, opened 2017-08-05
fix: Force support multidex on all configurations

### LINKED ISSUE #417: Force use of support multidex on all configurations
Currently test and main app use different versions of the dependency creating problems in build

iamareebjamal:
Fixes #417 

- Add support multidex lib
- Force apply on all configurations

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -67,6 +67,12 @@ sentry {
     autoUpload false
 }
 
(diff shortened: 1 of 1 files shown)


iamareebjamal:
@mejariamol @shreyasnbhat @Shailesh351 Please review

PR #415, opened 2017-08-04
feat: Add threetenabp and refactor DateUtils

iamareebjamal:
Fixes #413 

- Add threetenabp
- Add rigorous tests

> Note to self. Date Handling in Java is difficult

PR #422, opened 2017-08-05
feat: Add date time pickers in create ticket layout

### LINKED ISSUE #419: Add DateTime picker to ticket create layout
A date time picker should be present in the ticket create layout for setting sales start and end time

iamareebjamal:
To be merged after #415 

- Add date/time pickers in layout
- Create Date bindings
- Add observable fields in ticket
- Verify dates in presenter

Fixes #419

screenshot-1501941266301
screenshot-1501941277419
screenshot-1501941288798
screenshot-1501941300952

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/app/common/app/binding/DateBindings.java
+++ b/app/src/main/java/org/fossasia/openevent/app/common/app/binding/DateBindings.java
@@ -1,0 +1,66 @@
+package org.fossasia.openevent.app.common.app.binding;
+
+import android.app.AlertDialog;
+import android.app.DatePickerDialog;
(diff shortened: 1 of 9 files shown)


mejariamol:
@iamareebjamal 
Can we make this change in this PR itself as we won't require creating another issue for it:
1. Add click on the edit fields itself and show the formatted date time in the fields.

iamareebjamal:
No, we need to show the ISO field to the organiser.Some may want to directly manipulate the string rather than fiddle with the picker as it is a lot quicker. Also, the binding with ticket's fields is with ISO date, if we show formatted date there, the server will reject it

PR #424, opened 2017-08-05
fix: Make HostSelectionInterceptor Singleton

### LINKED ISSUE #423: Dynamic API url not working
Dynamic API URL not changing the base URL at the time of login/signup

iamareebjamal:
- Make DI graph provide Retrofit instead of Builder
- Fix presenter base url change API

Fixes #423


mejariamol:
@iamareebjamal where did we actually lose it?

iamareebjamal:
When the explicit provides annotation was removed for @inject annotation on the constructor. Also, please do review while I am away. I can answer your questions when I am back, but it'll be a waste of time for both of us as PRs could have been merged by now

PR #429, opened 2017-08-06
refactor: Add checkstyle, PMD and FindBugs

### LINKED ISSUE #428: Add static analysers and styling checks for Project
Add static analysers:

- Checkstyle
- Findbugs
- PMD

with custom project rules and make travis check them at PR

iamareebjamal:
Fixes #428 

- Add checkstyle, PMD and findbugs dependencies
  - Fix checkstyle violations
  - Update dependencies in ReadMe
- Fix findbugs violations
  - Add findbugs annotations
  - Add findbugs exclude filters
  - Add annotations on respective methods
  - Update Travis build file for check
- Improve checkstyle rules
  - Add customised rules for checkstyle
  - Increase difficulty level
  - Fix new violations
- Add PMD rules and finalize
  - Add customized PMD rules
  - Fix PMD violations
  - Change travis script to build instead of check
  - Finalize changes and violations

### DIFF
--- a/app/config/findbugs.xml
+++ b/app/config/findbugs.xml
@@ -1,0 +1,39 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+
+<!--
+FindBugs exclude filter
+Severity: HARD
+-->
+<FindBugsFilter>
+    <Match>
+        <Class name="~.*\.BuildConfig" />
+    </Match>
+    <Match>
+        <Class name="~.*\.R" />
+    </Match>
+    <Match>
+        <Class name="~.*\.R\$.*" />
+    </Match>
+
+    <!-- Ignore generated database class -->
+    <Match>
+        <Class name="~.*_Table" />
+    </Match>
+
+    <!-- Ignore static write for Context -->
+    <Match>
+        <Class name="org.fossasia.openevent.app.OrgaApplication" />
+        <Bug pattern="ST_WRITE_TO_STATIC_FROM_INSTANCE_METHOD " />
+    </Match>
+    <Match>
+        <Class name="org.fossasia.openevent.app.OrgaProvider" />
+        <Bug pattern="ST_WRITE_TO_STATIC_FROM_INSTANCE_METHOD " />
+    </Match>
+
+
+    <!-- Don't complain about mockito in tests. -->
+    <Match>
+        <Field name="~.*mockitoRule"/>
+        <Bug pattern="URF_UNREAD_PUBLIC_OR_PROTECTED_FIELD" />
+    </Match>
+</FindBugsFilter>
--- a/app/src/main/java/org/fossasia/openevent/app/common/app/binding/BindingAdapters.java
+++ b/app/src/main/java/org/fossasia/openevent/app/common/app/binding/BindingAdapters.java
@@ -83,9 +83,9 @@ public final class BindingAdapters {
         }
     }
 
-    @BindingAdapter(value = { "imageUrl", "placeholder" }, requireAll = false)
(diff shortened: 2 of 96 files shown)


iamareebjamal:
@mejariamol Please review

app/src/main/java/org/fossasia/openevent/app/common/app/binding/BindingAdapters.java
```diff
@@ -27,7 +27,11 @@
 import org.fossasia.openevent.app.common.utils.ui.CircleTransform;
 import org.fossasia.openevent.app.module.main.MainActivity;
 
-public class BindingAdapters {
+public final class BindingAdapters {
+
+    private BindingAdapters() {
```
mejariamol: why all these empty constructors?

iamareebjamal: https://stackoverflow.com/questions/14398747/hide-utility-class-constructor-utility-classes-should-not-have-a-public-or-def

https://stackoverflow.com/questions/32375149/is-it-mandatory-utility-class-should-be-final-and-private-constructor

app/config/findbugs.xml
```diff
@@ -0,0 +1,39 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+
+<!--
+FindBugs exclude filter
+Severity: HARD
+-->
+<FindBugsFilter>
```
mejariamol: can we add data binding generated files here too?

iamareebjamal: We don't need to test for the files we did not write

mejariamol: files added here to filter are auto generated ones right?

iamareebjamal: Yes, the ones which are always present, not the ones which are generated on annotation processing

mejariamol: data binding classes are auto generated and are present always. please take a look at this https://github.com/andriydruk/RxDNSSD/blob/master/library/findbugs-filter.xml  @iamareebjamal

iamareebjamal: He has just added those out of paranoia rather than facts. As you can see already, the build hasn't failed. Meaning that find bugs is not parsing those files at all

mejariamol: >the build hasn't failed. Meaning that find bugs is not parsing those files at all

this is used to exclude these files from checking syntax bugs right? @iamareebjamal

iamareebjamal: Yes, but these won't be checked anyways

mejariamol: Build fails when it doesn't fine some id in the data binding. Which means build checks data binding files too.

iamareebjamal: Is your build failing? Cuz neither Travis's is nor mine

mejariamol: No, not now. I want to say, build checks data binding files too. Remember the `checkedIn()` issue when refactoring to nextgen..

Just to confirm the file names added here will be excluded for syntax check right? If yes then we should include data binding generated files too. As we don't edit them.

iamareebjamal: That was other issue. It was related to lombok. FindBugs is not picking data binding files. Or else, it would definitely show errors. This may be because I have used a custom plugin to load it. Maybe it excludes them automatically. Besides  have removed DataBinding files from checkstyle which was actually checking them. I mean to say I have checked it. DataBinding files are not getting checked

iamareebjamal:
@niranjan94 Please merge

PR #447, opened 2017-08-09
fix: removes default date from the nav header

### LINKED ISSUE #446: Remove default date shown in nav header
when no event is selected, a default date is shown in the nav head which should not be.

mejariamol:
- checks nullability of the event start date in the attendee bottom sheet layout
- also hides soft keyboard on auth button click

fixes #446


mejariamol:
@iamareebjamal please review

iamareebjamal:
@mejariamol It's correct. Will approve one build fix is merged

PR #448, opened 2017-08-10
feat: Add Job Scheduler

### LINKED ISSUE #404: Local act and global sync for offline first app implementation
As the app is coming to final stages, this crucial feature needs to be implemented ASAP. All the network based actions should be dispatched to a JobScheduler Job if they are failed due to network error or timeout and be configured to be completed on a network connection with exponential backoff strategy.

The persistence of Jobs can be handled by the JobScheduler itself, but we also need to show the user what jobs are pending, which will need an attribute in the table for attendee (check in) and ticket for status.

The task can be broken into 2 parts:
- [ ] Scheduling of Attendee Check In
- [ ] Scheduling of Ticket creation

Canceling of pending jobs can be handled in a later issue

There are two options for JobScheduler which are API version agnostic : 
https://github.com/evernote/android-job
https://github.com/firebase/firebase-jobdispatcher-android

The first one is preferred as it does not require Google Play Services installed on the device. @niranjan94 Please confirm if this constraint is valid

The issue is open for now

iamareebjamal:
- Add JobScheduler, JobCreator and Attendee Check in job
- Make subsequent changes in check in presenter
- Make model state persistent and update UI accordingly
- Fix minor bugs in DatabaseChangeListener
- Create ObservableBoolean converter
- Add tests for above

Fixes #404


iamareebjamal:
@mejariamol @Shailesh351 @heysadboy Please review

mejariamol:
@iamareebjamal
There needs a slight modification in the current approach. Can be handled in a new issue.
Currently on attendee check in request, if network is not available the checking field is set to true and the job is added to the queue. But in the mean while attendee's local status is unclear until the job is done. 

Expected Approach should be:
Update the local status of the attendee when a job is added to the queue. A separate section shall be maintained to display the pending jobs in the app. But locally nothing should be pending. that is why we are giving offline support. On an attendee check in, initially the list of pending jobs and if found it is updated to the new request or if not found a new job is added. 

@iamareebjamal what do you say?

iamareebjamal:
If we update the status, then it may be misleading to the organizer who may think it is already updated. Pending is showing essentially that the job is not complete, even if there is a connection. The organizer should know that actually this is not updated yet and is pending because if not, and the request fails, we'll have to go modify the attendee status again to original one and this can create a lot of problems and inconsistencies from the database. There can be several reasons for failures, e.g, if organizer clicks check in twice and only one request fails, it will be impossible to determine the actual state of the attendee

Surely there needs to be a job queue for handling job cancellation

mejariamol:
>If we update the status, then it may be misleading to the organizer who may think it is already updated

Organizer need to know the status of the real attendee at the event not at the server that is our part. 

>and the request fails, we'll have to go modify the attendee status again to original one and this can create a lot of problems

on job failure the job should be re added to the queue. No need to modify the local status. As local status is what the actual status of the attendee. The status at the server is not. 

For example. If an attendee A is checked in using the app when network was down so job was added to the queue. But the attendee can't wait at the entry till the job is done. He shall be passed inside. Thats why we need to update the attendee's local status which is important and add the job to the queue until its not completed.

@iamareebjamal @niranjan94

iamareebjamal:
If the request never succeeds? The app will show wrong attendee status. If our server was as instantaneous as firebase, we could do this as we'd get the response of failure ASAP but here there can be several causes for it.

Most of the time organizer will want to know the actual status on the server and will be okay to see that the request is queued and will be synced when the network is back on. 

Another thing which can be done is that we do not toggle the state but show the intermediate state like a green ring on requested check in and `Pending Check In` to show like the clock icon chat messengers show

mejariamol:
@iamareebjamal Shall I create separate issue for this which can be taken up once this is merged?

iamareebjamal:
Please create another issue and start on it :smile:

PR #449, opened 2017-08-11
fix: emit data always from the local database

### LINKED ISSUE #435: Ticket type value hides from attendee on swipe refresh
Attendees page
In the response for `getAttendees` request from the server, all the other fields of the ticket are null excluding id. Hence ticket type is not shown after swipe refresh. Whereas it is shown when attendees are loaded from the local database.

mejariamol:
- on forced data request, load data from the server and update the local database and emit from the local database
- build method of AbstractObservableBuilder is modified to do the same.
- fixes missing fields in the emitted data due to relationships with ids in the models.

fixes #435


mejariamol:
@iamareebjamal please review.

mejariamol:
Yeah. looking into that. Actually mocking behavior of database repository is creating an issue.

mejariamol:
@iamareebjamal please review.

PR #456, opened 2017-08-12
feat: Add event detail activity

### LINKED ISSUE #453: Event Information Activity
An event info activity should be present to show basic event info

iamareebjamal:
- Add about event activity, fragment, presenter
- Change equals for CircleTransform
- Relax some checking rules

Fixes #453

screenshot-1502557656859
screenshot-1502557667016


iamareebjamal:
@mejariamol Please review

mejariamol:
@iamareebjamal 
Why separate activity for it? I think it should be hosted in MainActivity itself.
And we already showing event image in the nav head.

iamareebjamal:
How can we show this in MainActivity?

mejariamol:
>How can we show this in MainActivity?

Only the content which is shown in 2nd screen shot. Event Image is not required to show in the about page separately as we are showing it already in the nav header.

iamareebjamal:
But how will we show this info in MainActivity? MainActivity is just a host for fragments.

About the image, the screen also shows logo. I obviously showed the image because it was an activity and collapsing toolbar could be used. Also, option for editing the event can be added in future in this activity. 

Detail views are generally activities

mejariamol:
>But how will we show this info in MainActivity? MainActivity is just a host for fragments.

I am saying host AboutFragment in MainActivity.

iamareebjamal:
I think detail views should be shown in separate activities. Because navigationally, they are not flat with the hierarchy. Like track -> track detail should not be in same frame (activity) and some isolated functionalities just related to event info can be handled through this. This is a standard in material guidelines to show a detail view with a collapsing toolbar. Using collapsing toolbar in Fragments is not easy and the stack management will be disrupted due to it. Same with the User Detail

It should be handled in another activity as well. This will also make it easy to create master detail in future because the natural process for it is - mobile -> detail activity | tablet -> detail fragment

iamareebjamal:
@mejariamol Please review

app/src/main/res/values/strings.xml
```diff
@@ -75,5 +75,9 @@
     <string name="sale_start_time">Sale start time</string>
     <string name="sale_end_time">Sale end time</string>
     <string name="pick">Pick :</string>
+    <string name="about_event">About Event</string>
```
mejariamol: Just `About` is fine. As all the menus are in the event context already

app/src/main/res/layout/about_event_detail.xml
```diff
@@ -0,0 +1,140 @@
...
+                    android:layout_height="wrap_content"
+                    android:orientation="vertical">
+
+                    <TextView
+                        android:layout_width="match_parent"
+                        android:layout_height="wrap_content"
+                        android:padding="@dimen/spacing_small"
+                        android:text="@string/details"
+                        android:background="#dedede"
+                        android:textAppearance="@style/TextAppearance.AppCompat.Display1" />
```
mejariamol: decrease this text size. You may want to use `textSize` instead.
We should keep the same text sizes for titles and body on the pages. take textSize of titles from event dashboard page.

iamareebjamal: No, using textSize is not recommended. These specific styles are provided in Material Guidelines so they are standard. And these are card titles, so they are different from normal titles

https://material.io/guidelines/style/typography.html#typography-styles

Contrarily, wherever textSize is being used, these should be used instead

iamareebjamal:
@mejariamol Please review now

Issue #452, opened 2017-08-12
Enhance Attendees Page


mejariamol:
Expected:
- [x] filter by pending sync

Working on it.

mejariamol:
Just realized that there is no need to have a separate page to show pending attendees check ins. A filter on attendees page itself will do the work. Updated the issue.

PR #459, opened 2017-08-12
feat: Enhance Attendees Page

mejariamol:
- Added filter by pending sync

fixing #452 

screenshot_2017-08-16-00-55-16

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/app/module/attendee/list/AttendeesFragment.java
+++ b/app/src/main/java/org/fossasia/openevent/app/module/attendee/list/AttendeesFragment.java
@@ -71,6 +71,8 @@ public class AttendeesFragment extends BaseFragment<IAttendeesPresenter> impleme
 
     private boolean initialized;
 
+    private static final String FILTER_SYNC = "FILTER_SYNC";
+
     public AttendeesFragment() {
         OrgaApplication
             .getAppComponent()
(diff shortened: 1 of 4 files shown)


mejariamol:
@iamareebjamal please review.

app/src/main/java/org/fossasia/openevent/app/module/attendee/list/AttendeesFragment.java
```diff
@@ -107,13 +110,36 @@ public void onCreate(Bundle savedInstanceState) {
...
+        switch (item.getItemId()) {
+            case R.id.pendingSync:
+                item.setChecked(!item.isChecked());
+                if (item.isChecked()) {
+                    item.getIcon().setTint(utilModel.getResourceColor(R.color.color_accent));
+                    fastItemAdapter.filter(IS_SYNC_PENDING);
+                } else {
+                    item.getIcon().setTint(Color.BLACK);
+                    fastItemAdapter.filter("");
+                }
```
iamareebjamal: Rather than doing this, add a sort popup dialog which will contain items by which the attendees can be sorted. It will only contain one entry for now -> Pending

In future, it can contain other's like attendees with paid tickets only, or checked in attendees / checked out attendees, etc

mejariamol:
@iamareebjamal please review.

app/src/main/res/menu/menu_attendees.xml
```diff
@@ -1,9 +1,23 @@
...
+
+    <item android:id="@+id/filter"
+        android:title="@string/filter"
+        android:icon="@drawable/ic_filter"
+        app:showAsAction="ifRoom">
+        <menu>
+            <item android:id="@+id/filterByNone"
+                android:title="@string/filter_by_none" />
+            <item android:id="@+id/filterBySync"
+                android:title="@string/filter_by_sync" />
```
iamareebjamal: These should be checkable filters, where they can be checked to filter by one criteria or more. And I guess by current implementation, you can either search or filter, not both? This should not happen. Filters should cascade along with search

mejariamol: that will be out of the scope of this PR. This PR uses fast adapter's filter. For that to happen we will have to filter the attendees separately. which can be taken up in new PR.

iamareebjamal: @mejariamol actually it just needs an AND operator be added in the fast adapter filter method

PR #460, opened 2017-08-13
feat: Add organizer detail

### LINKED ISSUE #454: User Profile Activity
A user profile activity should be present to show basic user info

iamareebjamal:
- Add organizer detail activity, fragment, presenter

Fixes #454

screenshot-1502616495025
screenshot-1502616473233
screenshot-1502616485558


iamareebjamal:
@mejariamol Please review last commit

app/src/main/res/layout/organizer_detail_layout.xml
```diff
@@ -0,0 +1,172 @@
...
+                    android:layout_height="wrap_content"
+                    android:orientation="vertical">
+
+                    <TextView
+                        android:layout_width="match_parent"
+                        android:layout_height="wrap_content"
+                        android:padding="@dimen/spacing_small"
+                        android:text="@string/details"
+                        android:background="#dedede"
+                        android:textAppearance="@style/TextAppearance.AppCompat.Headline"
```
mejariamol: Same here too. We should maintain same text sizes pattern on all the pages. These title are bit larger. Keep same title size as on dashboard page.

iamareebjamal:
@mejariamol That can lead to overlap in some cases

app/src/main/res/layout/organizer_detail_layout.xml
iamareebjamal: This is the same as in event detail page. These are card headers

mejariamol: But these seem little bigger. I think these title sizes at least should be smaller than the toolbar title size.

iamareebjamal: No, that is not true at all. Then what'd be the point of any typography sizes larger than it in the design guidelines?

mejariamol: I am saying in this context. Here title in the toolbar is organizer name. And the titles in the content are subtitles (telling information about organizer)

iamareebjamal: Every activity has a toolbar and can contain larger TextViews. "In certain scenarios, use the larger subheading style instead of the smaller Body style. Some of those scenarios include when information is presented as small snippets or when titles are paired with lines of Body-styled text."

Because the title has to be always 20sp - "Across form factors, text that appears in the app bar should use the Title style, Medium 20sp."

mejariamol: >Some of those scenarios include when information is presented as small snippets or when titles are paired with lines of Body-styled text."

:+1:

iamareebjamal:
@mejariamol They are the only ones available. Perhaps you can point me to the solid ones which are not present

PR #463, opened 2017-08-13
fix: Make auth token required for every request

### LINKED ISSUE #461: Some requests are made without auth token
Current workflow only adds auth token if the original request comes without 401 error. Events API does not return 401 for published events. This makes organizer not able to access all of his events

Working on it

iamareebjamal:
- Replace selective authenticator with definite interceptor

Fixes #461


iamareebjamal:
@mejariamol Please review

iamareebjamal:
@mejariamol Please review

PR #473, opened 2017-08-14
refactor: Move unrelated model methods to delegates

### LINKED ISSUE #296: Create DTO and move delegate attributes from models
Our model classes are getting bloated by logic and computed properties that should not be present in a data class. 

Examples:
- Attendee
- Event

Create DTO (Data Transfer Objects) implementing excess logic and to encapsulate models, and use them instead to handle the computed properties

iamareebjamal:
- Create delegate interfaces and classes
- Move all unrelated model methods to them
- Use lombok delegate to generate methods
- Fix layout and tests

Fixes #296


iamareebjamal:
@mejariamol Please see now

iamareebjamal:
@mejariamol Please see now

PR #478, opened 2017-08-16
feat: Make detail views refreshable

### LINKED ISSUE #477: Make detail views refreshable
Currently, the detail views of event and user are not refreshable.

Working on it

iamareebjamal:
Fixes #477

- Make detail views refreshable
- Add new composite Rx Transformer
- Disable findbugs fail on error

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -77,6 +77,9 @@ configurations.all {
 
 check {
     abortOnError true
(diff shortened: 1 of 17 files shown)


mejariamol:
>This will not be used in fragment reloading

yeah right. That issue won't be here.

PR #487, opened 2017-08-19
fix: Modify placeholders in the Nav Header

### LINKED ISSUE #486: Replace Nav Header Placeholder image to without text
Currently the placeholder image in the Nav Header is with text which is overlapping with other texts shown in the header like organizer name. 

Expected: Change the image to no text.

Working on it.

Current View:

screenshot_2017-08-19-17-04-55

mejariamol:
- user placeholder icon is replaced and the position is changed
- event image is replaced

fixes #486

screenshot_2017-08-21-20-42-24


mejariamol:
@iamareebjamal how is this? I think this is good for now.Custom image can be added later. please review.

iamareebjamal:
@mejariamol We don't need raster icons, please. If they are changed in future, they'll pollute history as they are binary files. Same can be achieved with vector icons. 

Also, the account name is very much below from the icon, please fix it

mejariamol:
>We don't need raster icons

The vector was not scalable. It was blur with the given width height.

>he account name is very much below from the icon, please fix it

I checked gmail app and material guideline, they have aligned image at top and name detail at bottom. I can increase the bottom margin for name little. What do you say? Or just below the image?

mejariamol:
and the icon used is same which was used earlier just in png format with better quality. The white is the background set.

iamareebjamal:
But we don't want to use raster. They are dpi dependent and size consuming. Vectors look the same on all devices, are small in size and are not binary files. For all non complex resources, it is recommended to use vectors

iamareebjamal:
> The vector was not scalable. It was blur with the given width height.

That's another issue. It can be fixed and should be. Raster is not the solution

The image does not appear to be that far away in this and is in middle

screenshot-1503215288502

mejariamol:
@iamareebjamal please review now. :smile: converted that one into xml.

iamareebjamal:
@mejariamol Looks cool. I am searching/creating a default header bg. We'll update both in the same PR. Will send it by night

iamareebjamal:
header


@mejariamol Header Image

mejariamol:
@iamareebjamal good one!

mejariamol:
@iamareebjamal please review.

PR #495, opened 2017-08-22
fix: Don't reload the page on selected the same menu item

### LINKED ISSUE #494: Reloads page data on same menu item selection
Page reloads data even when the same menu item is selected.

Expected:
Do nothing, just close the drawer on same menu item selection.

Working on it.

mejariamol:
- added case to check the selected menu item id with last one in loadFragment method

fixes #494


mejariamol:
@iamareebjamal please review.

PR #521, opened 2017-08-28
fix: default color to the event detail palette

### LINKED ISSUE #520: Add default colors to event detail palette
Default colors to the event detail palette were loosed with #493 .

Expected:
When nav header image is default placeholder, set event detail palette color to grey and text color to white.

taking this up.

mejariamol:
- added background attribute to the container in the XML layout

fixes #520

screenshot_2017-08-28-17-21-16


mejariamol:
@iamareebjamal please review.

PR #533, opened 2017-09-06
chore: Ignore apk files

### LINKED ISSUE #532: Ignore apk files from git
Currently, gitgnore does not ignore built release apk files

iamareebjamal:
Fixes #532 

- Add app/*apk to gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,4 @@ UserInterfaceState.xcuserstate
 .externalNativeBuild
 
 /app/build
+*apk


.gitignore
```diff
@@ -32,3 +32,4 @@ UserInterfaceState.xcuserstate
 .externalNativeBuild
 
 /app/build
+/app/*apk
```
mejariamol: just `*.apk` would be better. What do you say? @iamareebjamal

Issue #548, opened 2017-09-23
Implement forgot password feature


ujjwalagrawal17:
**Expected Behaviour**

we need to implement forgot password function.

**Would you like to work on the issue?**

yes

Masquerade0097:
@iamareebjamal Can i work on this issue?

iamareebjamal:
Yes @Masquerade0097

srv-twry:
@iamareebjamal Can i take it since it seems @Masquerade0097 isn't working on it ?

Masquerade0097:
@srv-twry I'm working on it.

shubham-padia:
This exists already

Masquerade0097:
@iamareebjamal  This feature is now implemented in the app but was not auto closed. Please close this.

iamareebjamal:
Fixed by #580

PR #574, opened 2018-01-07
feature: Recovery token request to email

### LINKED ISSUE #573: Request recovery token on email
**Is it a ?**

- [ ] Bug Report

- [x] Feature Request
    
- [ ] Chore
    

**Feature Description**

In case user forgot his password, for resettng his password firstly the user need to enter his email id on which the server will send a recovery token, which will furthur be used for resetting the password. 

**Expected Behaviour**
Nothing available till now.

   
**Description**
Forgot Password module will be created in auth module which contains the fragment thorugh which user can input his email id and request a token on it. The server will send the token on the same only if the user is already registered on the site ie. he/she already has a login email id, password.

Yes

**Parent Issue**
#548

Masquerade0097:
Fixes #573 
Parent Issue #548 

Changes:
1. Created Forgot password Fragment
2. User inputs email id and after clicking Request Token button the recovery token will be sent to the user's email id.

Screenshots for the change: 
scr1
scr2

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/app/common/data/models/CustomObjectWrapper.java
+++ b/app/src/main/java/org/fossasia/openevent/app/common/data/models/CustomObjectWrapper.java
@@ -1,0 +1,19 @@
+package org.fossasia.openevent.app.common.data.models;
+
+import java.util.Collections;
+import java.util.Map;
+
+/*
+ * This class helps to wrap java objects with custom objects such that users can define
+ * their own tags to wrap objects with while serialising in JSON.
+*/
+
+public final class CustomObjectWrapper {
+
+    private CustomObjectWrapper() {
+    }
+
+    public static <E> Map<String, E> withLabel(String outerWrap, E wrappedObject) {
+        return Collections.singletonMap(outerWrap, wrappedObject);
+    }
+}
--- a/app/src/main/java/org/fossasia/openevent/app/common/data/network/EventService.java
+++ b/app/src/main/java/org/fossasia/openevent/app/common/data/network/EventService.java
@@ -2,12 +2,15 @@ package org.fossasia.openevent.app.common.data.network;
 
 import org.fossasia.openevent.app.common.data.models.Attendee;
 import org.fossasia.openevent.app.common.data.models.Event;
+import org.fossasia.openevent.app.common.data.models.RequestToken;
+import org.fossasia.openevent.app.common.data.models.RequestTokenResponse;
 import org.fossasia.openevent.app.common.data.models.Ticket;
 import org.fossasia.openevent.app.common.data.models.User;
 import org.fossasia.openevent.app.common.data.models.dto.Login;
 import org.fossasia.openevent.app.common.data.models.dto.LoginResponse;
 
 import java.util.List;
+import java.util.Map;
 
 import io.reactivex.Completable;
 import io.reactivex.Observable;
@@ -17,6 +20,10 @@ import retrofit2.http.GET;
 import retrofit2.http.PATCH;
 import retrofit2.http.POST;
 import retrofit2.http.Path;
+
+/* TODO: The current API doesn't conform to JSONAPI always. Revert to JSONAPI
+ implementation once this is fixed.
+ */
 
 public interface EventService {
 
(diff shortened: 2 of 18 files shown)


Masquerade0097:
@iamareebjamal Please review.

Masquerade0097:
@iamareebjamal Please review.

iamareebjamal:
@Masquerade0097 The build is failing due to checkstyle violations. Please fix it so that we can review and approve in one go and directly merge :smile:

app/src/main/java/org/fossasia/openevent/app/common/data/models/CustomObjectWrapper.java
```diff
@@ -0,0 +1,15 @@
...
+/**
+ * Created by priyanshu on 8/1/18.
+ */
+
+public final class CustomObjectWrapper {
+
+    public static <E> Map<String, E> withLabel(String outerWrap, E wrappedObject) {
+        return Collections.singletonMap(outerWrap, wrappedObject);
+    }
+}
```
iamareebjamal: Please remove author comments

app/src/main/java/org/fossasia/openevent/app/common/data/network/EventService.java
```diff
@@ -56,4 +59,6 @@
     @PATCH("attendees/{attendee_id}?include=ticket,event,order&fields[event]=id&fields[ticket]=id")
     Observable<Attendee> patchAttendee(@Path("attendee_id") long attendeeId, @Body Attendee attendee);
 
+    @POST("auth/reset-password")
+    Observable<RequestTokenResponse> requestToken(@Body Map<String,RequestToken> reqToken);
```
iamareebjamal: Add space after comma

app/src/main/java/org/fossasia/openevent/app/common/data/models/CustomObjectWrapper.java
iamareebjamal: Also, why is this used?

Masquerade0097: The token Json request has only email in Json body wrapped by defauly root element data. So to wrap the RequestToken POJO by data element i have to use a custom wrapper otherwise it wasn't getting wrapped. I couldn't use SerializationFeature.WRAP_ROOT_ELEMENT on ObjectWrapper because there was only one instance of it and it was mapped to all the mapped classses ie changing it would have effect on all the mapped classes.

iamareebjamal: That's not the reason actually. You have to use `jsonapi-converter` for the model. See other models like event.

Masquerade0097: Using josnapi-converter, 3 tags come by default - type, id, attributes and the body gets wrapped by attributes tag. I've tried to modify it but can't get it to the required form - 
{"data":{"type":"token","id":"0","attributes":{"email":"foxvixen@fox.com"}}}

I've come up with 2 ways of doing this - 
1. With CustomObjectWrapper I'm able to wrap RequestToken object with data tag and serialise it. 2. By extending the ObjectMapper class and changing it to wrap those objects with the root element tag ie. "data" which have a particular annotation on them like JsonRootName("data") but i think first one's better.

iamareebjamal: OK @Masquerade0097 seems like the problem is that API does not conform to JSONAPI always. Can you please add a TODO message in the API interface to revert to JSONAPI implementation once this is fixed. Also please add a comment on CustomDataWrapper about why it is used.

The implementation of your PR is of extremely good quality, thanks a lot :smile:

Masquerade0097: Ya, I'll add the comment and TODO. Also I've written some tests for the files which i have made or modified so should I squash them in this commit or make a separate commit for that.

iamareebjamal: Both ways are fine

Masquerade0097:
@iamareebjamal Please review

PR #596, opened 2018-01-23
feat: Added option to Create Event

### LINKED ISSUE #518: Add option to create Event
Add option to create an event with required fields

This issue follows creation of event with only basic bare minimum fields

Masquerade0097:
Fixes #518 

Changes: User can now create Event in the app with mandatory fields.

Screenshots for the change: 
brscr2

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/app/common/data/models/Event.java
+++ b/app/src/main/java/org/fossasia/openevent/app/common/data/models/Event.java
@@ -1,16 +1,10 @@
 package org.fossasia.openevent.app.common.data.models;
 
-import android.databinding.ObservableBoolean;
-
-import com.fasterxml.jackson.annotation.JsonAutoDetect;
 import com.fasterxml.jackson.annotation.JsonIgnore;
-import com.fasterxml.jackson.annotation.JsonInclude;
-import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.PropertyNamingStrategy;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonNaming;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
-import com.fasterxml.jackson.databind.ser.std.ToStringSerializer;
 import com.github.jasminb.jsonapi.LongIdHandler;
 import com.github.jasminb.jsonapi.annotations.Id;
 import com.github.jasminb.jsonapi.annotations.Relationship;
@@ -29,7 +23,6 @@ import org.fossasia.openevent.app.common.data.models.dto.ObservableString;
 import org.fossasia.openevent.app.common.utils.json.ObservableStringDeserializer;
 import org.fossasia.openevent.app.common.utils.json.ObservableStringSerializer;
 
-import java.io.File;
 import java.util.List;
 
 import lombok.AllArgsConstructor;
@@ -44,18 +37,10 @@ import lombok.experimental.Delegate;
 @AllArgsConstructor
 @JsonNaming(PropertyNamingStrategy.KebabCaseStrategy.class)
 @Table(database = OrgaDatabase.class, allFields = true)
-@JsonInclude(JsonInclude.Include.NON_DEFAULT)
 @EqualsAndHashCode(exclude = { "eventDelegate", "analytics" })
 @SuppressWarnings({ "PMD.ExcessivePublicCount", "PMD.TooManyFields" })
-@JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY, getterVisibility = JsonAutoDetect.Visibility.NONE,
-    setterVisibility = JsonAutoDetect.Visibility.NONE, creatorVisibility = JsonAutoDetect.Visibility.NONE,
-    isGetterVisibility = JsonAutoDetect.Visibility.NONE)
 public class Event implements Comparable<Event>, IHeaderProvider {
 
-    private static final String DEFAULT_STRING = "example";
-    private static final String DEFAULT_PROTOCOL = "http:";
-    private static final String DEFAULT_URL = "www.example.com";
-    private static final String DEFAULT_CURRENCY = "INR";
     public static final String STATE_DRAFT = "draft";
     public static final String STATE_PUBLISHED = "published";
 
@@ -98,39 +83,19 @@ public class Event implements Comparable<Event>, IHeaderProvider {
     public String bankDetails;
     public String chequeDetails;
     public String identifier;
-
     public boolean isComplete;
-
-    @JsonSerialize(using = ToStringSerializer.class)
     public double latitude;
-    @JsonSerialize(using = ToStringSerializer.class)
     public double longitude;
-    @JsonSerialize(using = ToStringSerializer.class)
     public boolean canPayByStripe;
-    @JsonSerialize(using = ToStringSerializer.class)
     public boolean canPayByCheque;
-    @JsonSerialize(using = ToStringSerializer.class)
     public boolean canPayByBank;
-    @JsonSerialize(using = ToStringSerializer.class)
     public boolean canPayByPaypal;
-    @JsonSerialize(using = ToStringSerializer.class)
     public boolean canPayOnsite;
-    @JsonSerialize(using = ToStringSerializer.class)
     public boolean isSponsorsEnabled;
-    @JsonSerialize(using = ToStringSerializer.class)
     public boolean hasOrganizerInfo;
-
-    @JsonProperty("is-sessions-speakers-enabled")
-    @JsonSerialize(using = ToStringSerializer.class)
     public boolean isSessionsSpeakersEnabled;
-    @JsonProperty("is-ticketing-enabled")
-    @JsonSerialize(using = ToStringSerializer.class)
     public boolean isTicketingEnabled;
-    @JsonProperty("is-tax-enabled")
-    @JsonSerialize(using = ToStringSerializer.class)
     public boolean isTaxEnabled;
-    @JsonProperty("is-map-shown")
-    @JsonSerialize(using = ToStringSerializer.class)
     public boolean isMapShown;
 
     @JsonSerialize(using = ObservableStringSerializer.class)
@@ -139,11 +104,6 @@ public class Event implements Comparable<Event>, IHeaderProvider {
     @JsonSerialize(using = ObservableStringSerializer.class)
     @JsonDeserialize(using = ObservableStringDeserializer.class)
     public ObservableString endsAt = new ObservableString();
-
-    @JsonIgnore
-    public final ObservableBoolean creating = new ObservableBoolean();
-    @JsonIgnore
-    public final ObservableBoolean deleting = new ObservableBoolean();
 
     @ColumnIgnore
     @Relationship("tickets")
@@ -160,26 +120,5 @@ public class Event implements Comparable<Event>, IHeaderProvider {
 
     private interface Excluding {
         List<Ticket> getEventTickets();
-    }
-
-    public void setDefaults() {
-        paymentCountry = DEFAULT_STRING;
-        paypalEmail = DEFAULT_STRING;
-        paymentCurrency = DEFAULT_CURRENCY;
-        organizerDescription = DEFAULT_STRING;
-        onsiteDetails = DEFAULT_STRING;
-        organizerName = DEFAULT_STRING;
-        locationName = DEFAULT_STRING;
-        privacy = DEFAULT_STRING;
-        codeOfConduct = DEFAULT_STRING;
-        state = STATE_DRAFT;
-        searchableLocationName = DEFAULT_STRING;
-        description = DEFAULT_STRING;
-        bankDetails = DEFAULT_STRING;
-        chequeDetails = DEFAULT_STRING;
-        originalImageUrl = DEFAULT_PROTOCOL + File.separator + File.separator + DEFAULT_URL;
-        externalEventUrl = DEFAULT_PROTOCOL + File.separator + File.separator + DEFAULT_URL;
-        logoUrl = DEFAULT_PROTOCOL + File.separator + File.separator + DEFAULT_URL;
-        ticketUrl = DEFAULT_PROTOCOL + File.separator + File.separator + DEFAULT_URL;
     }
 }
--- a/app/src/main/java/org/fossasia/openevent/app/common/data/models/Event.java
+++ b/app/src/main/java/org/fossasia/openevent/app/common/data/models/Event.java
@@ -1,8 +1,16 @@
 package org.fossasia.openevent.app.common.data.models;
 
+import android.databinding.ObservableBoolean;
+
+import com.fasterxml.jackson.annotation.JsonAutoDetect;
 import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.PropertyNamingStrategy;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonNaming;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import com.fasterxml.jackson.databind.ser.std.ToStringSerializer;
 import com.github.jasminb.jsonapi.LongIdHandler;
 import com.github.jasminb.jsonapi.annotations.Id;
 import com.github.jasminb.jsonapi.annotations.Relationship;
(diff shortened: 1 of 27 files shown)


app/src/main/java/org/fossasia/openevent/app/common/data/models/CreateEvent.java
```diff
@@ -0,0 +1,113 @@
...
+@AllArgsConstructor
+@NoArgsConstructor
+@JsonNaming(PropertyNamingStrategy.KebabCaseStrategy.class)
+@SuppressWarnings({ "PMD.TooManyFields" })
+@JsonInclude(JsonInclude.Include.ALWAYS)
+@EqualsAndHashCode(exclude = { "creating", "deleting" })
+@JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY, getterVisibility = JsonAutoDetect.Visibility.NONE,
+    setterVisibility = JsonAutoDetect.Visibility.NONE, creatorVisibility = JsonAutoDetect.Visibility.NONE,
+    isGetterVisibility = JsonAutoDetect.Visibility.NONE)
+public class CreateEvent {
```
iamareebjamal: We don't need another model. Original model can be reused

app/src/main/java/org/fossasia/openevent/app/common/data/models/CreateEventWrap.java
```diff
@@ -0,0 +1,20 @@
...
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+
+import lombok.Builder;
+import lombok.Data;
+
+@Data
+@Builder
+@JsonInclude(JsonInclude.Include.ALWAYS)
+public class CreateEventWrap {
```
iamareebjamal: Same here. Please see the format of creating tickets for more info

app/src/main/java/org/fossasia/openevent/app/common/data/models/CreateEvent.java
Masquerade0097: Original model Event structure is different from the requested one, it contains many unnecessary fields and also I had to assign default values for all the fields. That's why I created another object.

iamareebjamal: This is duplicating code, we'll have to maintain two models this way. Please use the existing one. Because having extra fields poses no problem in POST request. For default fields, one static method can be created to populate those when creating a new one

Masquerade0097: @iamareebjamal   I've tried using the Event object already present but it contains id tag when JSON request is made which I'm not able to remove because it gives HTTP 400 error - `"Event id not found"` . Is there any way I can hide that id tag( `@JsonIgnore` doesn't work for it, neither does `@JsonInclude(JsonInclude.Include.NON_NULL` when setting id to null)  like currently it's being done in Ticket object. In the Ticket create request, the id tag is not present although it is present in the Object, i can't understand how it's done.

iamareebjamal: Null fields should not be read by server. Can you please post the request being generated

Masquerade0097: When id is set to null, the request doesn't generate. The event create activity crashes.

Masquerade0097: id field is not setting to null by default. This is the request JSON -
 `D/OkHttp: {"data":{"type":"event","id":"0","attributes":{"analytics":{"checkedIn":{},"donationTickets":{},"freeTickets":{},"paidTickets":{},"soldDonationTickets":{},"soldFreeTickets":{},"soldPaidTickets":{},"totalAttendees":{},"totalSale":{},"totalTickets":{}},"bank-details":"example","can-pay-by-bank":"false","can-pay-by-cheque":"false","can-pay-by-paypal":"false","can-pay-by-stripe":"false","can-pay-onsite":"false","cheque-details":"example","code-of-conduct":"example","description":"example","ends-at":"2018-01-30T16:54:56.956+05:30","event-delegate":{"header":"LIVE","headerId":2337004},"external-event-url":"http://www.example.com","has-organizer-info":"false","is-complete":false,"is-sponsors-enabled":"false","latitude":"0.0","location-name":"example","logo-url":"http://www.example.com","longitude":"0.0","name":"I fgyhuh","onsite-details":"example","organizer-description":"example","organizer-name":"example","original-image-url":"http://www.example.com","payment-country":"example","payment-currency":"INR","paypal-email":"example","privacy":"example","searchable-location-name":"example","starts-at":"2018-01-24T16:54:56.956+05:30","state":"draft","ticket-url":"http://www.example.com","timezone":"Asia/Kolkata","is-map-shown":"false","is-sessions-speakers-enabled":"false","is-tax-enabled":"false","is-ticketing-enabled":"false"}}}`

iamareebjamal: Set to null

Masquerade0097:
@iamareebjamal Please review.

app/src/main/java/org/fossasia/openevent/app/common/data/models/Event.java
```diff
@@ -32,10 +44,18 @@
 @AllArgsConstructor
 @JsonNaming(PropertyNamingStrategy.KebabCaseStrategy.class)
 @Table(database = OrgaDatabase.class, allFields = true)
+@JsonInclude(JsonInclude.Include.NON_DEFAULT)
 @EqualsAndHashCode(exclude = { "eventDelegate", "analytics" })
 @SuppressWarnings({ "PMD.ExcessivePublicCount", "PMD.TooManyFields" })
+@JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY, getterVisibility = JsonAutoDetect.Visibility.NONE,
```
iamareebjamal: Why is this needed?

app/src/main/java/org/fossasia/openevent/app/common/data/models/Event.java
```diff
@@ -32,10 +44,18 @@
 @AllArgsConstructor
 @JsonNaming(PropertyNamingStrategy.KebabCaseStrategy.class)
 @Table(database = OrgaDatabase.class, allFields = true)
+@JsonInclude(JsonInclude.Include.NON_DEFAULT)
```
iamareebjamal: Why is this needed?

app/src/main/java/org/fossasia/openevent/app/common/data/models/Event.java
```diff
@@ -47,55 +67,84 @@
...
+    public boolean isSessionsSpeakersEnabled;
+    @JsonProperty("is-ticketing-enabled")
+    @JsonSerialize(using = ToStringSerializer.class)
+    public boolean isTicketingEnabled;
+    @JsonProperty("is-tax-enabled")
+    @JsonSerialize(using = ToStringSerializer.class)
+    public boolean isTaxEnabled;
+    @JsonProperty("is-map-shown")
+    @JsonSerialize(using = ToStringSerializer.class)
+    public boolean isMapShown;
```
iamareebjamal: This shouldn't be serialized to string. If this is a requirement, it is more likely a bug on server

app/src/main/java/org/fossasia/openevent/app/common/data/models/Event.java
```diff
@@ -112,4 +161,25 @@ public Event() { }
...
+        state = STATE_DRAFT;
+        searchableLocationName = DEFAULT_STRING;
+        description = DEFAULT_STRING;
+        bankDetails = DEFAULT_STRING;
+        chequeDetails = DEFAULT_STRING;
+        originalImageUrl = DEFAULT_PROTOCOL + File.separator + File.separator + DEFAULT_URL;
+        externalEventUrl = DEFAULT_PROTOCOL + File.separator + File.separator + DEFAULT_URL;
+        logoUrl = DEFAULT_PROTOCOL + File.separator + File.separator + DEFAULT_URL;
+        ticketUrl = DEFAULT_PROTOCOL + File.separator + File.separator + DEFAULT_URL;
+    }
```
iamareebjamal: By single responsibility principle, this should not be in this class at all, this class is model which should be mostly untouched

iamareebjamal: Also, don't initialize by default. These fields should be null if not provided by user

app/src/main/java/org/fossasia/openevent/app/common/data/models/Event.java
```diff
@@ -32,10 +44,18 @@
...
 @SuppressWarnings({ "PMD.ExcessivePublicCount", "PMD.TooManyFields" })
+@JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY, getterVisibility = JsonAutoDetect.Visibility.NONE,
+    setterVisibility = JsonAutoDetect.Visibility.NONE, creatorVisibility = JsonAutoDetect.Visibility.NONE,
+    isGetterVisibility = JsonAutoDetect.Visibility.NONE)
 public class Event implements Comparable<Event>, IHeaderProvider {
 
+    private static final String DEFAULT_STRING = "example";
+    private static final String DEFAULT_PROTOCOL = "http:";
+    private static final String DEFAULT_URL = "www.example.com";
+    private static final String DEFAULT_CURRENCY = "INR";
```
iamareebjamal: See comment below about single responsibility

app/src/main/java/org/fossasia/openevent/app/module/event/create/CreateEventPresenter.java
```diff
@@ -0,0 +1,74 @@
...
+    @Inject
+    public CreateEventPresenter(IEventRepository eventRepository) {
+        this.eventRepository = eventRepository;
+        LocalDateTime current = LocalDateTime.now();
+
+        String isoDate = DateUtils.formatDateToIso(current);
+        event.getStartsAt().set(isoDate);
+        event.getEndsAt().set(isoDate);
+        event.setDefaults();
+        event.setId(null);
```
iamareebjamal: It's null by default

app/src/main/java/org/fossasia/openevent/app/module/event/create/CreateEventPresenter.java
```diff
@@ -0,0 +1,74 @@
...
+    public void createEvent() {
+        if (!verify())
+            return;
+
+        eventRepository
+            .createEvent(event)
+            .compose(dispose(getDisposable()))
+            .compose(erroneous(getView()))
+            .doOnSubscribe(disposable -> event.creating.set(true))
+            .doFinally(() -> event.creating.set(false))
```
iamareebjamal: Rather than using event observable, use Rx transformers to do this and the view should be already implementing correct methods to show or hide the progress

app/src/main/java/org/fossasia/openevent/app/common/data/models/Event.java
Masquerade0097: This was added to reject the fields with non default values such that the request json doesn't contain unwanted stuff but it don't affect the functionality of the app i.e. event can be created without it. So i'll remove it.

app/src/main/java/org/fossasia/openevent/app/common/data/models/Event.java
Masquerade0097: There are some fields like `isMapShown` which by default are getting serialized to two fields - `is-map-shown` and `map-shown`. So to prevent this I added the above annotation after which it as only getting serialized to `is-map-shown`.

iamareebjamal: Yes

app/src/main/java/org/fossasia/openevent/app/common/data/models/Event.java
Masquerade0097: Yes I have to add them. Please see my above comment.

app/src/main/java/org/fossasia/openevent/app/common/data/models/Event.java
Masquerade0097: When the fields are not initialized by default they are having null values which in turn causes NullPointerException in Dashboard which in turn shows blank. Also just after creating event I get a whte screen and app hangs though the event is created successfully. But on choosing that event in the event list layout, again the app hangs.  This is the error log for NullPointerException - 
`E/Logger: An exception occurred : value is null
          java.lang.NullPointerException: value is null
              at io.reactivex.internal.functions.ObjectHelper.requireNonNull(ObjectHelper.java:39)
              at io.reactivex.Single.just(Single.java:655)
              at org.fossasia.openevent.app.common.utils.core.CurrencyUtils.getCurrencySymbol(CurrencyUtils.java:44)
              at org.fossasia.openevent.app.module.main.MainPresenter.lambda$start$1$MainPresenter(MainPresenter.java:68)
              at org.fossasia.openevent.app.module.main.MainPresenter$$Lambda$2.accept(Unknown Source)
              at io.reactivex.internal.observers.LambdaObserver.onNext(LambdaObserver.java:63)
              at io.reactivex.internal.operators.observable.ObservableDoOnEach$DoOnEachObserver.onNext(ObservableDoOnEach.java:103)
              at io.reactivex.internal.operators.observable.ObservableDoOnEach$DoOnEachObserver.onNext(ObservableDoOnEach.java:103)
              at io.reactivex.internal.observers.DisposableLambdaObserver.onNext(DisposableLambdaObserver.java:58)
              at io.reactivex.subjects.PublishSubject$PublishDisposable.onNext(PublishSubject.java:264)
              at io.reactivex.subjects.PublishSubject.onNext(PublishSubject.java:183)
              at org.fossasia.openevent.app.common.data.Bus.pushSelectedEvent(Bus.java:19)
              at org.fossasia.openevent.app.module.event.list.EventsListAdapter$EventRecyclerViewHolder.lambda$new$0$EventsListAdapter$EventRecyclerViewHolder(EventsListAdapter.java:76)
              at org.fossasia.openevent.app.module.event.list.EventsListAdapter$EventRecyclerViewHolder$$Lambda$0.onClick(Unknown Source)
              at android.view.View.performClick(View.java:5612)
              at android.view.View$PerformClick.run(View.java:22285)
              at android.os.Handler.handleCallback(Handler.java:751)
              at android.os.Handler.dispatchMessage(Handler.java:95)
              at android.os.Looper.loop(Looper.java:154)
              at android.app.ActivityThread.main(ActivityThread.java:6123)
              at java.lang.reflect.Method.invoke(Native Method)
              at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:867)
              at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:757)`

app/src/main/java/org/fossasia/openevent/app/module/event/create/CreateEventPresenter.java
Masquerade0097: Ya, I'll remove it.

app/src/main/java/org/fossasia/openevent/app/common/data/models/Event.java
Masquerade0097: Please read my comment below about setting default values. If needed, I'll initialize them in the presenter.

app/src/main/java/org/fossasia/openevent/app/common/data/models/Event.java
Masquerade0097: Earlier i was getting server error sucht that it was accpting `"false"` and not `false` that's why i had to add it but now I'm not getting any related errors. So I'll remove it then.

app/src/main/java/org/fossasia/openevent/app/module/event/create/CreateEventPresenter.java
Masquerade0097: Ok

app/src/main/java/org/fossasia/openevent/app/common/data/models/Event.java
iamareebjamal: Then ask for the payment currency from the user. And please create an issue to handle null currency type

Masquerade0097:
@iamareebjamal I've made the changes, please have a look.

Masquerade0097:
@iamareebjamal I've made the changes, please have a look.

app/src/main/java/org/fossasia/openevent/app/common/data/models/Event.java
```diff
@@ -47,54 +52,58 @@
...
+    public boolean isTicketingEnabled;
+    public boolean isTaxEnabled;
+    public boolean isMapShown;
+
+    @JsonSerialize(using = ObservableStringSerializer.class)
+    @JsonDeserialize(using = ObservableStringDeserializer.class)
+    public ObservableString startsAt = new ObservableString();
+    @JsonSerialize(using = ObservableStringSerializer.class)
+    @JsonDeserialize(using = ObservableStringDeserializer.class)
+    public ObservableString endsAt = new ObservableString();
```
iamareebjamal: This is a deprecated method of binding and warnings are displayed. Data binding developers will remove this method soon, they have advised to not use it

Masquerade0097: @iamareebjamal CreateEvent as well as AboutEvent uses this method of binding and gives a warning. Both use the same DateService methods etc can I open a separate issue for this warning and use the same mechanism for both of them because changing them in one place will give error in other. What do you say?

iamareebjamal: Yup :+1:

PR #607, opened 2018-01-30
fix: Event List UI improvement

### LINKED ISSUE #599: UI
**Is it a ?**

- [ ] Bug Report

- [ ] Feature Request
    
- [ ] Chore
        

**Actual Behaviour**

The Draft Textview doesn't adjust accordingly with the length of Event name.

**Expected Behaviour**

State here what the feature should enable the user to do.

**Steps to reproduce it**

Login -> Select Events in Drawer  

**Screenshots of the issue**

escr1

Moto G4 Plus
    
Yes

Masquerade0097:
Fixes #599 

Changes: Event List layout is compatible with big as well as small event names. Draft TextView doesn't get truncated.

Screenshots for the change: 
dscr1


app/src/main/res/layout/event_layout.xml
```diff
@@ -44,39 +44,53 @@
...
-                android:text="@{event.name}"
-                android:textSize="@dimen/text_size_normal"
-                android:textColor="@android:color/black"
-                android:paddingEnd="@dimen/spacing_medium"
-                android:paddingLeft="@dimen/spacing_medium"
-                android:paddingRight="@dimen/spacing_medium"
-                android:paddingStart="@dimen/spacing_medium"/>
-
-            <FrameLayout
+            <RelativeLayout
```
iamareebjamal: RelativeLayout is not needed for this job. What feature is it providing that a linearlayout may not

Masquerade0097: If you require LinearLayout then I'll change it. It's just that Relative Layout does only one extra measure than LinearLayout and since the hierarchy isn't much complex so it would not cause any performance related issue. Anyways no problem, I'll change it.

iamareebjamal: Actually we'll later want to use constraint layout, and it is easier to migrate this way

Masquerade0097: @iamareebjamal  Currently, Constraint Layouts are used in very few layout files. Should I create an issue for migrating the existing layouts to constraint layouts?

Issue #624, opened 2018-02-02
App crashes when value in Quantity textview is erased in create new ticket form.


AkshayCHD:
**Is it a ?**

- [x] Bug Report

- [ ] Feature Request
    
- [ ] Chore
    
### Bug Report    

**Actual Behaviour**

App crashes when value in Quantity edittext of New ticket form is erased

**Expected Behaviour**

App should not crash

**LogCat for the issue**

02-02 22:24:09.995 3281-3281/org.fossasia.eventyay E/MessageQueue-JNI: Exception in MessageQueue callback: handleReceiveCallback
02-02 22:24:10.095 3281-3281/org.fossasia.eventyay E/MessageQueue-JNI: java.lang.NumberFormatException: Invalid long: ""
                                                                           at java.lang.Long.invalidLong(Long.java:124)
                                                                           at java.lang.Long.parseLong(Long.java:345)
                                                                           at java.lang.Long.parseLong(Long.java:321)
                                                                           at org.fossasia.openevent.app.common.app.binding.BindingAdapters.strToLong(BindingAdapters.java:45)
                                                                           at org.fossasia.openevent.app.databinding.TicketCreateFormBinding$5.onChange(TicketCreateFormBinding.java:184)
                                                                           at android.databinding.adapters.TextViewBindingAdapter$1.onTextChanged(TextViewBindingAdapter.java:367)
                                                                           at android.widget.TextView.sendOnTextChanged(TextView.java:7991)
                                                                           at android.widget.TextView.handleTextChanged(TextView.java:8053)
                                                                           at android.widget.TextView$ChangeWatcher.onTextChanged(TextView.java:10157)
                                                                           at android.text.SpannableStringBuilder.sendTextChanged(SpannableStringBuilder.java:1033)
                                                                           at android.text.SpannableStringBuilder.replace(SpannableStringBuilder.java:559)
                                                                           at android.text.SpannableStringBuilder.delete(SpannableStringBuilder.java:225)
                                                                           at android.text.SpannableStringBuilder.delete(SpannableStringBuilder.java:224)
                                                                           at android.text.method.BaseKeyListener.backspaceOrForwardDelete(BaseKeyListener.java:106)
                                                                           at android.text.method.BaseKeyListener.backspace(BaseKeyListener.java:51)
                                                                           at android.text.method.BaseKeyListener.onKeyDown(BaseKeyListener.java:215)
[...]

PR #632, opened 2018-02-07
fix: Event Create Layout with all fields

### LINKED ISSUE #610: feat: Adding optional fields to Create Event Layout
**Is it a ?**

- [ ] Bug Report

- [x] Feature Request
    
- [ ] Chore
    
**Feature Description**

Modify the Create Event layout to request all the optional fields from the user as well.
   
Yes

Masquerade0097:
Fixes #610 

Changes: 

1. I've made ValidateUtils to validate url and email because the Ilhasoft data binding validator makes the field as a mandatory field on which it is applied. I've made a regex to verify the urls and emails such that the user gets error response while typing the links but the regex rejects only subset of all possible bad links rejected by server, for better accuracy I would need the regex used in the server. So I'm able to provide error feedback to the user on most of the wrong inputs, on others the server will give `HTTP 500` error.

2. I've used JsonProperty attribute for some fields in Event model because the fields with names like - `isSessionSponsorsEnabled`, `canPayByPaypal` weren't getting serialized to the required format on their own.

3. The issue #624 is present in longitude and latitude field which uses similar binding adapters methods on which the assigned person is working so I didn't correct it.

Screenshots for the change: 
bnscr1
bnscr2
bnscr3
bnscr4
bnscr5


Masquerade0097:
@iamareebjamal Please review.

iamareebjamal:
Amazing work. I'll be suggesting some improvements on JSON serialization soon (did you try kebab case strategy).

About setter methods, just make the model properties public, then databinding will be able to read them

Masquerade0097:
@iamareebjamal The model already uses kebabCaseStrategy, the problem in serialization was that some fields similar to canPayByPal weren't getting serialized at all and others similar to isSessionsSpeakersEnabled were getting serialized to - seesions-speakers-enabled but JsonProperty annotation resolved this.
Also all the model properties are already public but still data binding can't read it.

iamareebjamal:
We were already binding event properties in Data Binding, can you give an example of a property that can't be read?

Masquerade0097:
I was having this problem with the boolean fields and not with the string fields so on searching i just found out that lombok has a slight different method naming behaviour regarding getters and setters of boolean fields. So those setters are made arn't needed now, I'll update the PR soon.

Masquerade0097:
@iamareebjamal I've removed the setters.

app/src/main/java/org/fossasia/openevent/app/module/event/create/CreateEventFragment.java
```diff
@@ -65,6 +69,17 @@ public View onCreateView(LayoutInflater inflater, @Nullable ViewGroup container,
...
         });
 
+        ArrayAdapter<CharSequence> adapter = new ArrayAdapter<>(getActivity(), android.R.layout.simple_spinner_item);
+        adapter.setDropDownViewResource(android.R.layout.simple_spinner_dropdown_item);
+
+        ArrayList<String> currencyCodesArray = new ArrayList<>(CurrencyUtils.getCurrencyCodesList());
+
+        for (int i = 0; i < currencyCodesArray.size(); i++) {
+            adapter.add(currencyCodesArray.get(i));
+        }
```
iamareebjamal: There are two redundant operations happening here. First a list is returned by CurrencyUtils. Then a new list is as currencyCodesArray. Double the memory double the work. Thirdly, Then it is iterated one by one without an iterator but through an index.

Try `adapter.addAll(CurrencyUtils.getCurrencyCodeList())`

If doesn't work, try

```Java
for (String code : CurrencyUtils.getCurrencyCodesList())
    adapter.add(code);
```

iamareebjamal: Secondly, it'd be better if this is moved to the presenter and added lazily but has low priority. If you want, you may move it

app/src/main/java/org/fossasia/openevent/app/module/event/create/CreateEventPresenter.java
```diff
@@ -53,11 +58,100 @@ private boolean verify() {
         return true;
     }
 
+    private boolean findEmptyLink(String uri) {
+        if (uri == null)
+            return false;
+        else return !("".equals(uri));
+    }
```
iamareebjamal: Remove this method, simply use `StringUtils.isEmpty`

iamareebjamal: Or better even, create a new static method called `StringUtils.emptyToNull` which returns null if string is empty or string itself if it is not

app/src/main/java/org/fossasia/openevent/app/module/event/create/CreateEventPresenter.java
```diff
@@ -53,11 +58,100 @@ private boolean verify() {
...
+
+    private void resolveEmptyLinks() {
```
iamareebjamal: Rename to `nullifyEmptyFields`, and make it protected so that we can test this. Also pass event through parameter

app/src/main/java/org/fossasia/openevent/app/module/event/create/CreateEventPresenter.java
```diff
@@ -1,10 +1,16 @@
 package org.fossasia.openevent.app.module.event.create;
 
+import android.support.design.widget.TextInputLayout;
+import android.text.Editable;
+import android.text.TextUtils;
+import android.text.TextWatcher;
```
iamareebjamal: Remove any and every Android-related dependency from presenter

app/src/main/java/org/fossasia/openevent/app/module/event/create/CreateEventPresenter.java
```diff
@@ -53,11 +58,100 @@ private boolean verify() {
...
+                    textInputLayout.setErrorEnabled(false);
+                }
+            }
+
+            @Override
+            public void afterTextChanged(Editable editable) {
+                // Nothing here
+            }
+        });
+    }
```
iamareebjamal: If we can't unit test something in Presenter, it's of no use. This should be present in View

app/src/main/java/org/fossasia/openevent/app/module/event/create/CreateEventPresenter.java
```diff
@@ -53,11 +58,100 @@ private boolean verify() {
...
```
iamareebjamal: This should be present in View

Masquerade0097:
@iamareebjamal Please review.

app/src/main/java/org/fossasia/openevent/app/module/event/create/CreateEventFragment.java
```diff
@@ -73,6 +85,81 @@ public void onStart() {
...
+                    textInputLayout.setErrorEnabled(false);
+                }
+            }
+
+            @Override
+            public void afterTextChanged(Editable editable) {
+                // Nothing here
+            }
+        });
+    }
```
iamareebjamal: Very nice job! Just last change here, and we are done. These two tasks are essentially same here except two things, validation method and error text. This can be made in a single method which takes a method reference and a string parameter. Also, use strings.xml template to render the text instead

Masquerade0097: Changes Done !!

PR #639, opened 2018-02-14
fix: app does not crash while creating new ticket or event

### LINKED ISSUE #638: App crashes in create new ticket
**Is it a ?**

- [X] Bug Report

- [ ] Feature Request
    
- [ ] Chore
    
### Bug Report    

**Actual Behaviour**

In create new ticket if you change the date to an invalid format and then click on "INVALID DATE" or click on the fab to create the ticket,app crashes

**Expected Behaviour**

App should not crash

**Steps to reproduce it**

Create new ticket
Change the start date or end date to an invalid format
Click on the fab or on "INVALID DATE"

**LogCat for the issue**

`02-14 15:22:33.333 28032-28032/org.fossasia.openevent.app E/DateUtils: org.threeten.bp.format.DateTimeParseException: Text '2018-02-14T15:22:24.' could not be parsed at index 20
`

**Screenshots of the issue**
ezgif com-video-to-gif

Yes

nikit19:
Fixes #638 

Changes: app does not crash while creating new ticket or event

Screenshots for the change(UPDATED): 
ezgif com-video-to-gif 2

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/app/common/app/binding/DateBindings.java
+++ b/app/src/main/java/org/fossasia/openevent/app/common/app/binding/DateBindings.java
@@ -13,6 +13,9 @@ import org.threeten.bp.LocalDate;
 import org.threeten.bp.LocalDateTime;
 import org.threeten.bp.LocalTime;
 import org.threeten.bp.ZonedDateTime;
+import org.threeten.bp.format.DateTimeParseException;
+
+import timber.log.Timber;
 
 public final class DateBindings {
 
@@ -27,6 +30,7 @@ public final class DateBindings {
         button.setText(DateUtils.formatDateWithDefault(format, isoDate));
     }
 
+    @SuppressWarnings("PMD.DataflowAnomalyAnalysis")
     private static void bindTemporal(Button button, ObservableField<String> date, String format,
                                      Function<ZonedDateTime, AlertDialog> dialogProvider) {
         if (date == null)
@@ -36,7 +40,12 @@ public final class DateBindings {
         button.setText(DateUtils.formatDateWithDefault(format, isoDate));
 
         button.setOnClickListener(view -> {
-            ZonedDateTime zonedDateTime = DateUtils.getDate(isoDate);
+            ZonedDateTime zonedDateTime = ZonedDateTime.now();
+            try {
+                zonedDateTime = DateUtils.getDate(isoDate);
+            } catch (DateTimeParseException pe) {
+                Timber.e(pe);
+            }
             dialogProvider.apply(zonedDateTime).show();
         });
     }
--- a/app/src/main/java/org/fossasia/openevent/app/module/event/create/CreateEventPresenter.java
+++ b/app/src/main/java/org/fossasia/openevent/app/module/event/create/CreateEventPresenter.java
@@ -11,6 +11,7 @@ import org.fossasia.openevent.app.module.event.create.contract.ICreateEventPrese
 import org.fossasia.openevent.app.module.event.create.contract.ICreateEventView;
 import org.threeten.bp.LocalDateTime;
 import org.threeten.bp.ZonedDateTime;
+import org.threeten.bp.format.DateTimeParseException;
 
 import javax.inject.Inject;
 
@@ -43,15 +44,19 @@ public class CreateEventPresenter extends BasePresenter<ICreateEventView> implem
     }
 
     private boolean verify() {
-        ZonedDateTime start = DateUtils.getDate(event.getStartsAt().get());
-        ZonedDateTime end = DateUtils.getDate(event.getEndsAt().get());
+        try {
+            ZonedDateTime start = DateUtils.getDate(event.getStartsAt().get());
+            ZonedDateTime end = DateUtils.getDate(event.getEndsAt().get());
 
-        if (!end.isAfter(start)) {
-            getView().showError("End time should be after start time");
+            if (!end.isAfter(start)) {
+                getView().showError("End time should be after start time");
+                return false;
+            }
+            return true;
+        } catch (DateTimeParseException pe) {
+            getView().showError("Please enter date in correct format");
             return false;
         }
-
-        return true;
     }
 
     protected void nullifyEmptyFields(Event event) {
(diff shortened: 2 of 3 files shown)


app/src/main/java/org/fossasia/openevent/app/common/utils/core/DateUtils.java
```diff
@@ -67,7 +68,14 @@ public static String formatDateToIso(@NonNull LocalDateTime date) {
...
             return ZonedDateTime.now();
-        return ZonedDateTime.parse(isoDateString).withZoneSameInstant(getZoneId());
+        try {
+            exception = false;
+            return ZonedDateTime.parse(isoDateString).withZoneSameInstant(getZoneId());
+        } catch (DateTimeParseException pe) {
+            Timber.e(pe);
+            exception = true;
+            return ZonedDateTime.now();
+        }
```
iamareebjamal: We never handle exceptions in util methods. If a string is invalid, why should we return current date to the user?

app/src/main/java/org/fossasia/openevent/app/common/utils/core/DateUtils.java
```diff
@@ -67,7 +68,14 @@ public static String formatDateToIso(@NonNull LocalDateTime date) {
     public static ZonedDateTime getDate(@NonNull String isoDateString) {
         if (isoDateString == null)
             return ZonedDateTime.now();
-        return ZonedDateTime.parse(isoDateString).withZoneSameInstant(getZoneId());
+        try {
+            exception = false;
+            return ZonedDateTime.parse(isoDateString).withZoneSameInstant(getZoneId());
+        } catch (DateTimeParseException pe) {
+            Timber.e(pe);
+            exception = true;
```
iamareebjamal: We never ever add a state to pure function, there's a reason why the mehod is static and not in a class

app/src/main/java/org/fossasia/openevent/app/common/utils/core/DateUtils.java
```diff
@@ -78,17 +86,11 @@ public static String formatDate(@NonNull String format, @NonNull String isoDateS
 
     @NonNull
     public static String formatDateWithDefault(@NonNull String format, @NonNull String isoString, @NonNull String defaultString) {
-        try {
+        getDate(isoString);
```
iamareebjamal: I mean the return is completely ignored, double computation

app/src/main/java/org/fossasia/openevent/app/common/utils/core/DateUtils.java
```diff
@@ -78,17 +86,11 @@ public static String formatDate(@NonNull String format, @NonNull String isoDateS
...
+        if (!exception)
```
iamareebjamal: This will return default string after any parsing error if getDate is not called and as they are not part of same function, it is impossible to debug the implications of state change. Furthermore, I have not even considered synchronization issues in a parallel thread. This solution is dead ringer for countless more bugs than it solves

Recognize the error and solve it rather than patching unneeded things :smile:

nikit19: `return ZonedDateTime.parse(isoDateString).withZoneSameInstant(getZoneId());`
This line causes a DateTimeParse exception in getDate
So the only thing that I could think of was adding a try catch block.
If I don't return current date when a string is invalid, what should I return?
Other than the code written in getDate, the rest of the code is a workaround to pass the unit test

iamareebjamal: Tell me this. If you ask someone time and their clock is broken, should they tell you that their clock is broken or just tell you any time. The method **should** throw exception if the date is in wrong format and not return current time

The problem is that the date should never be in wrong format, so try to find out on what cases it is throwing exception and remove those cases from happening

nikit19: The date will be only be in wrong format when we manually change it in the editText(like it is being done in the gif)
Can making the editText non-editable be a solution?

iamareebjamal: Or you could try catch but in the calling method and not in the util method. Util method should always throw error in wrong format

app/src/main/java/org/fossasia/openevent/app/common/app/binding/DateBindings.java
```diff
@@ -36,7 +40,12 @@ private static void bindTemporal(Button button, ObservableField<String> date, St
         button.setText(DateUtils.formatDateWithDefault(format, isoDate));
 
         button.setOnClickListener(view -> {
-            ZonedDateTime zonedDateTime = DateUtils.getDate(isoDate);
+            ZonedDateTime zonedDateTime = ZonedDateTime.now();
+            try {
+                zonedDateTime = DateUtils.getDate(isoDate);
+            } catch (DateTimeParseException pe) {
+                Timber.e(pe);
+            }
```
iamareebjamal: Only apply if it is correctly parsed

app/src/main/java/org/fossasia/openevent/app/module/event/create/CreateEventPresenter.java
```diff
@@ -42,9 +43,17 @@ public Event getEvent() {
...
+        ZonedDateTime end;
+        try {
+            start = DateUtils.getDate(event.getStartsAt().get());
+            end = DateUtils.getDate(event.getEndsAt().get());
+        } catch (DateTimeParseException pe) {
+            getView().showError("Please enter date in correct format");
+            return false;
+        }
 
         if (!end.isAfter(start)) {
```
iamareebjamal: Move everything inside try block

app/src/main/java/org/fossasia/openevent/app/common/app/binding/DateBindings.java
nikit19: Actually if I do that the date picker dialog won't open when the date is invalid  and then there will be no way of correcting the date.
Done with other requested changes

app/src/main/java/org/fossasia/openevent/app/common/app/binding/DateBindings.java
```diff
@@ -27,6 +30,7 @@ private static void setPickedDate(LocalDateTime pickedDate, Button button, Strin
         button.setText(DateUtils.formatDateWithDefault(format, isoDate));
     }
 
+    @SuppressWarnings("PMD")
```
iamareebjamal: Specifically target the PMD warning

nikit19:
Done with the changes

PR #649, opened 2018-02-18
fix: Allign ticket's default sale end time with frontend

### LINKED ISSUE #614: Change sale end time default for Create Ticket Screen.
**Is it a ?**

- [x] Bug Report

- [ ] Feature Request
    
- [ ] Chore
    
### Bug Report    

**Actual Behaviour**

The Sale End time of the CreateTicket Screen starts at the current time.

**Expected Behaviour**

It would be better if it's default end time is the end time of the Event itself.
Not a major issue, can be labelled ```later```

**Steps to reproduce it**

Use CreateTicket screen to create a Ticket.

**LogCat for the issue**

NA

Yes.

srv-twry:
Fixes #614 

Changes: Alligned the ticket's default sale end time with the front end.
The front end adds 10 days to the current date for the default end time. If the number of days left for the event is less than 10 days then it is set to the end date of the event.

Screenshots for the change: 
videotogif_2018 02 18_21 32 29

@iamareebjamal Please review.

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/app/module/ticket/create/CreateTicketPresenter.java
+++ b/app/src/main/java/org/fossasia/openevent/app/module/ticket/create/CreateTicketPresenter.java
@@ -26,10 +26,17 @@ public class CreateTicketPresenter extends BasePresenter<ICreateTicketView> impl
     public CreateTicketPresenter(ITicketRepository ticketRepository) {
         this.ticketRepository = ticketRepository;
         LocalDateTime current = LocalDateTime.now();
+        String startDate = DateUtils.formatDateToIso(current);
+        ticket.getSalesStartsAt().set(startDate);
 
-        String isoDate = DateUtils.formatDateToIso(current);
-        ticket.getSalesStartsAt().set(isoDate);
-        ticket.getSalesEndsAt().set(isoDate);
+        LocalDateTime salesEndTime = current.plusDays(10);
+        LocalDateTime eventEndTime = DateUtils.getIsoOffsetTimeFromTimestamp(ContextManager.getSelectedEvent().getEndsAt().get());
+        //if less than 10 days are available in the event.
+        if (salesEndTime.isAfter(eventEndTime)) {
+            salesEndTime = eventEndTime;
+        }
+        String endDate = DateUtils.formatDateToIso(salesEndTime);
+        ticket.getSalesEndsAt().set(endDate);
         ticket.setType("free");
     }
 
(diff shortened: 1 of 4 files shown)


srv-twry:
Forgot to update the unit tests. Doing now.

app/src/main/java/org/fossasia/openevent/app/module/ticket/create/CreateTicketPresenter.java
```diff
@@ -26,10 +26,17 @@
...
         this.ticketRepository = ticketRepository;
         LocalDateTime current = LocalDateTime.now();
+        String startDate = DateUtils.formatDateToIso(current);
+        ticket.getSalesStartsAt().set(startDate);
 
-        String isoDate = DateUtils.formatDateToIso(current);
-        ticket.getSalesStartsAt().set(isoDate);
-        ticket.getSalesEndsAt().set(isoDate);
+        LocalDateTime salesEndTime = current.plusDays(10);
+        LocalDateTime eventEndTime = DateUtils.getIsoOffsetTimeFromTimestamp(ContextManager.getSelectedEvent().getEndsAt().get());
```
srv-twry: @iamareebjamal Can you please help here ? 
I am trying to make sure that the sales end time doesn't go beyond the event's end time. To do that i have to use 
```ContextManager```'s static method ```getSelectedEvent``` to get the event's end time. 
Now Mockito doesn't support static methods. That's why the build fails. How to go about doing this ? Should i use PowerMockito for testing ?

Comment (iamareebjamal) [reply app/src/main/java/org/fossasia/openevent/app/module/ticket/create/CreateTicketPresenter.java]: Use ContextManager's `setSelectedEvent`

File: app/src/test/java/org/fossasia/openevent/app/unit/presenter/TicketCreatePresenterTest.java [hunk 2]
```diff
@@ -40,10 +44,17 @@ public void setUp() {
         RxJavaPlugins.setIoSchedulerHandler(scheduler -> Schedulers.trampoline());
         RxAndroidPlugins.setInitMainThreadSchedulerHandler(schedulerCallable -> Schedulers.trampoline());
 
+        setupMockEvent();
+        ContextManager.setSelectedEvent(event);
```
iamareebjamal: Set to null after test

srv-twry:
Changes: Fixed the ```CreateTicketFragment``` test.

@iamareebjamal Please review.

PS: Couldn't find out the cause of failure of circleci build.

iamareebjamal:
Ignore CircleCI, it is unreliable right now

app/src/test/java/org/fossasia/openevent/app/unit/presenter/TicketCreatePresenterTest.java
iamareebjamal: You just need to set null in this method only, no need to add to tearDown

PR #687, opened 2018-03-02
feature: Implement local timezone selection by default while creating…

### LINKED ISSUE #683: Select local timezone by default while creating event
**Is it a ?**

- [ ] Bug Report

- [x] Feature Request
    
- [ ] Chore
    
### Feature Request

**Feature Description**

The local timezone should be fetched and displayed as default when creating an event

mishuvs:
Fixes #683

Changes: 
    1. added new method setDefaults() to ICreateEventView.java
    2. CreateEventFragment.java implements method setDefaults()
    3. start() in CreateEventPresenter.java calls setDefaults()

Screenshot of the change: 
screenshot_2018-03-02-06-35-38-834_org fossasia eventyay

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/app/module/event/create/CreateEventFragment.java
+++ b/app/src/main/java/org/fossasia/openevent/app/module/event/create/CreateEventFragment.java
@@ -24,6 +24,7 @@ import org.fossasia.openevent.app.databinding.EventCreateLayoutBinding;
 import org.fossasia.openevent.app.module.event.create.contract.ICreateEventPresenter;
 import org.fossasia.openevent.app.module.event.create.contract.ICreateEventView;
 
+import java.util.Arrays;
 import java.util.List;
 
 import javax.inject.Inject;
@@ -155,6 +156,16 @@ public class CreateEventFragment extends BaseBottomSheetFragment<ICreateEventPre
     @Override
     public void onSuccess(String message) {
         ViewUtils.showSnackbar(binding.getRoot(), message);
+    }
+
+    @Override
+    public List<String> getTimeZoneList() {
+        return Arrays.asList(getResources().getStringArray(R.array.timezones));
+    }
+
+    @Override
+    public void setDefaultTimeZone(int index) {
+        binding.form.timezoneSpinner.setSelection(index);
     }
 }
 
--- a/app/src/main/java/org/fossasia/openevent/app/module/event/create/CreateEventPresenter.java
+++ b/app/src/main/java/org/fossasia/openevent/app/module/event/create/CreateEventPresenter.java
@@ -13,6 +13,9 @@ import org.threeten.bp.LocalDateTime;
 import org.threeten.bp.ZonedDateTime;
 import org.threeten.bp.format.DateTimeParseException;
 
+import java.util.List;
+import java.util.TimeZone;
+
 import javax.inject.Inject;
 
 import static org.fossasia.openevent.app.common.app.rx.ViewTransformers.dispose;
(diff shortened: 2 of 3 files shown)


app/src/main/java/org/fossasia/openevent/app/module/event/create/CreateEventFragment.java
```diff
@@ -132,6 +133,22 @@ public void attachCurrencyCodesList(List<String> currencyCodesList) {
         binding.form.currencySpinner.setAdapter(adapter);
     }
 
+    @Override
+    public void setDefaults() {
```
iamareebjamal: No point of creating a method in Fragment when there is no information to be sent from Presenter.

Also, setDefaults is vague method name

Create a method in fragment called `getTimeZoneList` and call it in presenter to get the `List<String>` of timezones

Then operate on the list asynchronously and update the UI accordingly

For doing so there would be another method in fragment call `setDefaultTimezone` which will take a string as argument and apply `binding.form.timezoneSpinner.setSelection(index)`

mishuvs: Hi, @iamareebjamal Thanks for the review! 
I made the following changes, in my local repo as directed by you:
1. in fragment:
``` 
public List<String> getTimeZoneList(){
        return Arrays.asList(getResources().getStringArray(R.array.timezones));
    }
```
2. in fragment:
```
@Override
    public void setDefaultTimeZone(String zoneID) {
        //set default time zone:
        List<String> timeZoneList = getTimeZoneList();
        int index = timeZoneList.indexOf(zoneID);
        if(index!=-1) binding.form.timezoneSpinner.setSelection(index);
    }
```

3. in presenter:
```
@Override
public void start() {
        ...
        getView().setDefaultTimeZone(TimeZone.getDefault().getID());
    }
```

But I don't understand why we need to call `getTimeZoneList` in the presenter (we can call it from the fragment directly)? And, can't we put the `getTimeZoneList` method in `DateUtils.java`? Also, since the task is not heavy, do we really need to perform it asynchronously?

iamareebjamal: No that's not how you should do it. Get the list of timezones from fragment in the presenter and then  convert them into an list observable and find the index of default timezone. Once you do that, you have to call the method on view to set the form selection. The "work" should be done in presenter so that it can be tested. 

OR

Another solution is to just insert the default timezone on top of the list so it is available readily and set the default selection to 0. I think this is more easy and efficient

mishuvs: @iamareebjamal how about this?:

in fragment:
```
public List<String> getTimeZoneList(){
        return Arrays.asList(getResources().getStringArray(R.array.timezones));
    }
```
in fragment:
```
@Override
    public void setDefaultTimeZone(int index) {
        binding.form.timezoneSpinner.setSelection(index);
    }
```
in presenter:
```
@Override
public void start() {
        ...
        //set default timezone
        List<String> timeZoneList = ((CreateEventFragment) getView()).getTimeZoneList();
        getView().setDefaultTimeZone(
            timeZoneList.indexOf(TimeZone.getDefault().getID())
        );
    }
```

iamareebjamal: We don't want any casts in the presenter, that's why we are using interfaces, the interface should automatically contain the method `getTimeZoneList`

Otherwise, it looks good

mishuvs: Thanks for helping @iamareebjamal 
I have added the method to the interface and will now update the pull request.
But before I do that, can you guide me on why my build test is failing? From travis-ci log, I found that the `androidCheckstyle` task execution is failing.

iamareebjamal: Yes, you have to run `./gradlew androidCheckstyle` and check the generated HTML which will tell you some of the issues your code changes might have added

mishuvs:
@iamareebjamal I have made the changes to the PR. Can you please have a look?

PR #688, opened 2018-03-02
fix: Create Event Activity will finish after event is created

### LINKED ISSUE #613: CreateEventActivity should be finished after creating a new Event.
**Is it a ?**

- [x] Bug Report

- [ ] Feature Request
    
- [ ] Chore
    
### Bug Report    

**Actual Behaviour**

After creation of Event is successful, the CreateEventActivity still remains open.

**Expected Behaviour**

It's not a bug but i think it's not standard for it to remain active. In most of the standard applications, the create activities are finished after successful creation.
Not a major issue hence can be labelled ```later```

**Steps to reproduce it**

Create an event using the CreateEventActivity

Yes.

leeyh20:
Fixes #613 

Changes: Create Event Activity will finish and send an intent to the fragment that started that activity, which is EventListFragment. (However, EventListFragment's parent activity is MainActivity, thus the intent will first be passed to the MainActivity) EventListFragment will then know that creating the event was successful as its onActivityResult function will be called.

Edit: Now just shows a toast instead.

Screenshots for the change: 
In this GIF, we create an event named "lol"
createevent

@srv-twry @iamareebjamal Could you help me to review this?
However, EventListFragment does not automatically refresh after the event is created. I have tried calling the function `getPresenter().loadUserEvents(true)` inside onActivityResult to try to refresh the events list, but to no avail. Is there a way to refresh without the user having to swipe?

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/app/module/event/create/CreateEventFragment.java
+++ b/app/src/main/java/org/fossasia/openevent/app/module/event/create/CreateEventFragment.java
@@ -173,15 +173,5 @@ public class CreateEventFragment extends BaseBottomSheetFragment<ICreateEventPre
     public void setDefaultTimeZone(int index) {
         binding.form.timezoneSpinner.setSelection(index);
     }
-
-    @Override
-    public List<String> getTimeZoneList() {
-        return Arrays.asList(getResources().getStringArray(R.array.timezones));
-    }
-
-    @Override
-    public void setDefaultTimeZone(int index) {
-        binding.form.timezoneSpinner.setSelection(index);
-    }
 }
 
--- a/app/src/main/java/org/fossasia/openevent/app/module/event/create/CreateEventFragment.java
+++ b/app/src/main/java/org/fossasia/openevent/app/module/event/create/CreateEventFragment.java
@@ -13,6 +13,7 @@ import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
 import android.widget.ArrayAdapter;
+import android.widget.Toast;
 
 import org.fossasia.openevent.app.OrgaApplication;
 import org.fossasia.openevent.app.R;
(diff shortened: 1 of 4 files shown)


iamareebjamal:
No need to make it that complicated. Just close the activity once it's done. This can be done in the fragment using getActivity().finish()

leeyh20:
@iamareebjamal If I immediately closed the activity without sending any intent, I could not show the Snackbar with the message “Event Created Successfully” on EventListFragment. Is there a simpler way?

leeyh20:
@iamareebjamal If I immediately closed the activity without sending any intent, I could not show the Snackbar with the message “Event Created Successfully” on EventListFragment.

The reason why I used onActivityResult was to display the Snackbar message at `ViewUtils.showSnackbar(binding.getRoot(), message);`

 Is there a simpler way?

iamareebjamal:
Show a Toast for the event creation message instead :smile:

leeyh20:
Oh that sounds good! I'll try that instead!

app/src/main/java/org/fossasia/openevent/app/module/event/create/CreateEventFragment.java
```diff
@@ -156,9 +155,9 @@ public void showProgress(boolean show) {
 
     @Override
     public void onSuccess(String message) {
-        Intent intent = new Intent();
-        intent.putExtra("message", message);
-        getActivity().setResult(Activity.RESULT_OK, intent);
+        final int duration = Toast.LENGTH_SHORT;
+        Toast toast = Toast.makeText(getActivity().getApplicationContext(), message, duration);
+        toast.show();
```
iamareebjamal: Since we do not want a reference of the toast to work on, we can just do `Toast.makeText(getActivity(), message, Toast.LENGTH_SHORT).show()`

Also, no need to get Application Context

leeyh20:
@iamareebjamal Thank you for your help, the issue was simple after all ^^ll Is there a way to make EventListFragment refresh after the event is created?

leeyh20:
@iamareebjamal Oh I see that another issue has been created for it haha.

app/src/main/java/org/fossasia/openevent/app/module/event/create/CreateEventFragment.java
```diff
@@ -154,7 +155,8 @@ public void showProgress(boolean show) {
 
     @Override
     public void onSuccess(String message) {
-        ViewUtils.showSnackbar(binding.getRoot(), message);
+        Toast.makeText(getActivity(), message, Toast.LENGTH_SHORT).show();
+        getActivity().finish();
```
iamareebjamal: Rather than closing in `onSuccess` create a new method in view contract called `close` and call it in the presenter, so that we can test the behaviour

leeyh20:
@iamareebjamal Thanks for your help! Is this what you meant by creating a new method? 
I have also tried to create a new unit test for it, or it is too similar to `shouldShowSuccessOnCreated`?

iamareebjamal:
Please resolve conflicts

PR #704, opened 2018-03-04
fix: Toolbar goes behind the status bar on login screen #681

### LINKED ISSUE #681: Toolbar goes behind the status bar on login screen
**Is it a ?**

- [x] Bug Report

- [ ] Feature Request
    
- [ ] Chore
    
### Bug Report    

**Actual Behaviour**

When typing in any field, the toolbar goes behind the status. 

**Expected Behaviour**

It should stay stuck at the top

**Steps to reproduce it**

Type in any text field in login screen

**Screenshots of the issue**

screenshot-1519932995513

mishuvs:
Fixes #681 

Changes: Add `windowSoftInputMode="adjustResize"` to AndroidManifest.xml

Screenshots for the change: 
appbarvisible

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -22,7 +22,8 @@
         <activity
             android:name=".module.auth.AuthActivity"
             android:theme="@style/AppTheme.Main.Light"
-            android:label="${appName}">
+            android:label="${appName}"
+            android:windowSoftInputMode="adjustResize">
             <intent-filter>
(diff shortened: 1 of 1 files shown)


iamareebjamal:
The issue is with the special wrapping style on screens and the should not require this flag since it may mask future errors

mishuvs:
@iamareebjamal Do you mean that I need to make changes the layout file?

iamareebjamal:
Yes, layout or styles

PR #719, opened 2018-03-05
bug: Remove multiple instances of currencies from currency list (#718)

### LINKED ISSUE #718: Remove multiple instances of same surrency from currency list
**Is it a ?**

- [x] Bug Report

- [ ] Feature Request
    
- [ ] Chore
    
### Bug Report    

**Actual Behaviour**

There are multiple instances of the same currency.

**Expected Behaviour**

The should be only one instance of each currency, except the ones which occur prior in the list. 

Yes

AkshayCHD:
Fixes #718 
Changes: There were several instances of same currencies which have been removed now.


AkshayCHD:
@iamareebjamal please review

PR #720, opened 2018-03-06
fix: Remove redundancy in case of date-time selection in the event cr…

### LINKED ISSUE #707: Remove redundancy in case of date-time selection in the event creation form
**Is it a ?**

- [x] Bug Report

- [ ] Feature Request
    
- [ ] Chore
    
### Bug Report    

**Actual Behaviour**

Both DatePicker-TimePicker and EditText field is shown. This makes up as a redundant part in the UI, and makes the form look bigger and complex.

**Expected Behaviour**

Only the Pickers should be shown, as they are more user-friendly.

**Screenshots of the issue**

Where-ever possible attach a screenshot of the issue.

Yes

mishuvs:
…eation form (#707)

Fixes #707 

Changes: 
1. Added checkboxes to toggle the visibility of text input for date and time selection for sale start time and the sale end time. Either the input fields will be visible or the picker. By default, the checkbox will be checked so as to make the user use the date and time pickers, as they enhance UX.
2. Added Missing padding attribute for the second date-time field.

dtnew


iamareebjamal:
The checkbox should be present in the widget itself

mishuvs:
Hi @iamareebjamal Pardon, I did not understand.
In the widget? Do you mean below 'Date and Time' header?

iamareebjamal:
By widget, I meant in the layout of the date time, there should be a checkbox which triggers the edittext to be shown

mishuvs:
@iamareebjamal Do you imply creating two checkboxes for the Date and Time pickers in the same line? So that each of them could be triggered individually? I think, doing this would be increasing unnecessary UI elements. And even the fields would look squashed. If a user prefers text input, he would prefer it for both the fields. And switching is easy.

Or what do you think about this? Does it look fine?:
dtpicker

iamareebjamal:
The thing is that these are not only used in create event form. So, the idea is to create reusable components

mishuvs:
@iamareebjamal I have made changes. Please review.

iamareebjamal:
Please reupload the images

mishuvs:
@iamareebjamal I have updated! Please review.

iamareebjamal:
Not updated

mishuvs:
> The thing is that these are not only used in create event form. So, the idea is to create reusable components

Currently, only the `create_ticket_form.xml` is using the same date-time layout component. And even when other layouts use it, they will all use both 'starts-at' and 'ends-at' fields.
I think that makes it whole a reusable component.
Besides that, adding the 4 child views on the same line will probably not look good.
Would you like it changed?

iamareebjamal:
There will be at least 12 layouts which will use this field in future, we want to be future proof. We can improve the UI iteratively if it does not look good

mishuvs:
Got it! :thumbsup:

mishuvs:
@iamareebjamal Please check.

iamareebjamal:
Please update the screenshots

iamareebjamal:
Please move the checkbox in next line and make the text small and changed to (Use Picker)

mishuvs:
@iamareebjamal Updated! Please review.

mishuvs:
Are any further changes required?

iamareebjamal:
We should move the picker layout to its own file and ensure it is being used everywhere

mishuvs:
Can we keep these changes for now, and open a new issue for that?

iamareebjamal:
If this checkbox layout is moved to another file, how will it require changes in other layouts?

mishuvs:
I was thinking of breaking down the task into smaller commits. 
1. Presently, we have made changes to the code for only one of the views to show. 
2. Next, we move this code to a new layout and update `create_event_form.xml`. 
3. Then we change other layout files (like `create_ticket_form.xml`), which also take date-time input so that they include the layout as well.

iamareebjamal:
1 & 2 can be done in this PR only

mishuvs:
@iamareebjamal Please review.

app/src/main/AndroidManifest.xml
```diff
@@ -22,7 +22,8 @@
         <activity
             android:name=".module.auth.AuthActivity"
             android:theme="@style/AppTheme.Main.Light"
-            android:label="${appName}">
+            android:label="${appName}"
+            android:windowSoftInputMode="adjustResize">
```
iamareebjamal: Why this?

app/src/main/res/layout/end_time.xml
```diff
@@ -0,0 +1,88 @@
...
+            android:id="@+id/picker_end"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:gravity="center"
+            android:text="@string/use_date_time_picker"
+            android:textSize="@dimen/text_size_extra_small"
+            android:checked="true"/>
+    </LinearLayout>
+
+</layout>
```
iamareebjamal: We don't want a layout for end time. How can we reuse this?

We want a general layout for the date time picker

app/src/main/res/layout/start_time.xml
```diff
@@ -0,0 +1,86 @@
...
+        <CheckBox
+            android:id="@+id/picker_start"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:text="@string/use_date_time_picker"
+            android:textSize="@dimen/text_size_extra_small"
+            android:checked="true" />
+    </LinearLayout>
+
+</layout>
```
iamareebjamal: Same here

app/src/main/res/values/strings.xml
```diff
@@ -182,6 +182,9 @@
     <string name="create_faq">Create FAQ</string>
     <string name="question">Question</string>
     <string name="answer">Answer</string>
+    <string name="use_date_time_picker">Use Picker</string>
+    <string name="event_start_time">Starts at:</string>
+    <string name="event_end_time">Ends at:</string>
```
iamareebjamal: Name should be `starts_at` `ends_at`

app/src/main/AndroidManifest.xml
mishuvs: Sorry, it crept in. Will remove.

app/src/main/res/layout/end_time.xml
mishuvs: One has `app:date="@{ event.startsAt }"` and other `app:date="@{ event.startsAt }"`, and similar respective properties for start and end. Or do I need include both in one layout?

iamareebjamal: You just send a different date reference in each layout. Look how you can pass data in each layout

app/src/main/res/values/strings.xml
mishuvs: I have removed them as we can pass those too, using binding. And currently we're using `sale_start_time` and `sale_end_time` in `create_event_form.xml`.

iamareebjamal:
If the text is "starts at", please rename string name to starts_at

mishuvs:
There's no string that reads "starts at". No additional strings like that were added.

iamareebjamal:
You should create one or rename from `sale_starts_at`

If `sale_starts_at` says "Starts At", it should be renamed to `starts_at`

mishuvs:
`sale_starts_at` reads "Sale starts at". It is left untouched as it would be useful in `ticket_create_form.xml`, where "Starts at" would be ambiguous.
`starts_at` reading "Starts at:" is added and replaced in `event_create_form.xml`.

PR #787, opened 2018-03-18
Login UI Improved : Textview Added

### LINKED ISSUE #749: Remove "LogIn" from Action Bar title
- [x] Feature Request
    
### Feature Request

**Feature Description**

In the Login Activity, instead of showing 'Log In' in action bar title which is very obvious looking at the screen, we can remove the action bar and display a designed textview with "Open Event Orga App from Fossasia".

YashishDua:
Fixes #749 

Changes: Added a textview welcoming organizer and making landing page looks like it is an app for the organiser.

Updated: 

n

### DIFF
--- a/app/src/main/res/layout/login_fragment.xml
+++ b/app/src/main/res/layout/login_fragment.xml
@@ -29,6 +29,18 @@
                 android:paddingRight="@dimen/spacing_normal"
                 android:paddingStart="@dimen/spacing_normal">
 
+                <TextView
+                    android:textStyle="bold"
+                    android:textColor="@color/color_accent"
+                    android:textSize="@dimen/text_size_larger"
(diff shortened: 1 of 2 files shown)


iamareebjamal:
Please add screenshots

YashishDua:
Added. Review it!

iamareebjamal:
Looks bloated, we can have a thin font instead

YashishDua:
Changed. Review now!

iamareebjamal:
Should this be present in signup page as well?

YashishDua:
I don't think so! Because if he/she is login in, it means for sure an organizer is logging in. On the other hand, while signing up he/she is just a user. When signs up then becomes a organizer,

PR #799, opened 2018-03-20
Fix:Settings and other NavItem should be in backstack should lead to …

### LINKED ISSUE #682: Settings fragment should be in backstack
**Is it a ?**

- [x] Bug Report

- [ ] Feature Request
    
- [ ] Chore
    
### Bug Report    

When clicking back in settings section, the app shows toast to exit if back is pressed twice

**Expected Behaviour**

The settings fragment should be killed and app should return to the main screen on back press

Prashant-Chaurasia:
…dashboard.

Fixes #682  

Changes: [Earlier on pressing back button in a NavItemfragment was leading to exit the app.But this should not be the case.Only when a person is on the dashboard fragment then only one can exit the app. So, this is a fix where on pressing back button in any NavItem Fragment it leads to dashboard screen and now from this fragment only he can exit.]

Screenshots for the change: 
video2gif_20180320_111824
video2gif_20180320_112331


app/src/main/java/org/fossasia/openevent/app/module/main/MainActivity.java
```diff
@@ -94,9 +94,17 @@ protected void onStart() {
...
         if (binding.drawerLayout.isDrawerOpen(GravityCompat.START)) {
             binding.drawerLayout.closeDrawer(GravityCompat.START);
-        } else {
+        }
+        else if(lastSelectedNavItemId!=R.id.nav_dashboard) {
+            binding.navView.setCheckedItem(R.id.nav_dashboard);
+            lastSelectedNavItemId = R.id.nav_dashboard;
+            Fragment fragment = EventDashboardFragment.newInstance(eventId);
+            fragmentManager.beginTransaction().replace(R.id.fragment_container, fragment).commit();
+        }
```
iamareebjamal: This is not using fragment backstack at all, instead this is creating a new instance of DashBoardFragment which shouldn't be the case

Prashant-Chaurasia: Yes I did this deliberately because if we use addtobackstack() method to add fragments to the backstack then what will happen is suppose if we go to faqs and then from faqs to tickets, then both will be added to the stack and then pressing back would lead to faqs and then again pressing back will lead to the dashboard. I firstly implemented the backstack, but to jump from any fragment to dashboard we need to loop through each item and pop from the stack.I'll implement it using backstack if you want me to.

iamareebjamal: Thats easy to handle and standard. Everytime to add to backstack you just update a boolean field and check if we are already one level on backstack and only add it if we aren't same with the popping of backstack

Prashant-Chaurasia:
@iamareebjamal Please review this PR.

app/src/main/java/org/fossasia/openevent/app/common/utils/ui/BackPressHandler.java
```diff
@@ -14,12 +16,23 @@
     public BackPressHandler() { }
 
     public void onBackPressed(Activity activity, Runnable action) {
-        if (backPressed + BACK_PRESS_RESET_TIME > System.currentTimeMillis()) {
-            action.run();
-        } else {
-            Toast.makeText(activity, "Press again to exit", Toast.LENGTH_SHORT).show();
+        if(activity instanceof MainActivity){
```
iamareebjamal: Checking instance is bad. Java is statically typed language. This is dynamic behaviour. Always prefer static behaviour when possible and it is possible here

app/src/main/java/org/fossasia/openevent/app/common/utils/ui/BackPressHandler.java
```diff
@@ -14,12 +16,23 @@
...
+            }
+            backPressed = System.currentTimeMillis();
+        }
+        else{
+            if (backPressed + BACK_PRESS_RESET_TIME > System.currentTimeMillis()) {
+                action.run();
+            } else {
+                Toast.makeText(activity, "Press again to exit", Toast.LENGTH_SHORT).show();
+            }
+            backPressed = System.currentTimeMillis();
```
iamareebjamal: Follow standard Java style guideline

app/src/main/java/org/fossasia/openevent/app/module/main/MainActivity.java
```diff
@@ -94,9 +96,17 @@ protected void onStart() {
     public void onBackPressed() {
         if (binding.drawerLayout.isDrawerOpen(GravityCompat.START)) {
             binding.drawerLayout.closeDrawer(GravityCompat.START);
-        } else {
-            backPressHandler.onBackPressed(this, super::onBackPressed);
         }
+        else {
+            if (fragmentManager.findFragmentById(R.id.fragment_container) instanceof EventListFragment ||
+                fragmentManager.findFragmentById(R.id.fragment_container) instanceof EventDashboardFragment) {
+                backPressHandler.onBackPressed(this, super::onBackPressed);
```
iamareebjamal: Again, "checking instance" is poor solution. You can see yourself how highly dependent it is on the current structure of app. Code should never even know what the structure of app is. Anytime I change these fragments or add new ones, I'll have to change this code. Which has nothing to do with these fragments

app/src/main/java/org/fossasia/openevent/app/module/main/MainActivity.java
```diff
@@ -207,7 +217,18 @@ private void loadFragment(int navItemId) {
...
+            fragmentManager.beginTransaction().replace(R.id.fragment_container, fragment).addToBackStack("root_fragment").commit();
+        }
+        else if(fragmentManager.getBackStackEntryCount()==1){
+            fragmentManager.beginTransaction().replace(R.id.fragment_container, fragment).addToBackStack("").commit();
+        }
+        else {
+            fragmentManager.popBackStack("root_fragment",0);
+            fragmentManager.beginTransaction().replace(R.id.fragment_container, fragment).addToBackStack("").commit();
+        }
+
```
iamareebjamal: Follow standard Java style guidelines

app/src/main/java/org/fossasia/openevent/app/module/main/MainActivity.java
```diff
@@ -207,7 +217,18 @@ private void loadFragment(int navItemId) {
                 fragment = EventDashboardFragment.newInstance(eventId);
                 break;
         }
-        fragmentManager.beginTransaction().replace(R.id.fragment_container, fragment).commit();
+        if(fragmentManager.getBackStackEntryCount()==0){
```
iamareebjamal: When we have control over pushing and popping the state, we should know that what state we are in and not need to get the backstack entry count. This is bad because it means we have limited the capability of the stack to a constant where this is not the desired behaviour.

Prashant-Chaurasia:
@iamareebjamal Please review this. :smiley:

app/src/main/java/org/fossasia/openevent/app/module/main/MainActivity.java
```diff
@@ -207,7 +222,18 @@ private void loadFragment(int navItemId) {
                 fragment = EventDashboardFragment.newInstance(eventId);
                 break;
         }
-        fragmentManager.beginTransaction().replace(R.id.fragment_container, fragment).commit();
+
+        if(fragment_count>1){
+            fragmentManager.popBackStack("root_fragment",0);
+            fragment_count--;
+        }
```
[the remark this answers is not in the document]
iamareebjamal: We only need a boolean field because we won't go beyond 1 level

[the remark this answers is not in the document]
Prashant-Chaurasia: Yes, I totally agree with you, we won't go beyond 1 level. But here suppose we go from fragments A(dashboard) to B to C then pressing back would lead to A. But in the process firstly A is added followed by B and now as we move to C, B is removed so, that backstack is AC . So, we need to keep a counter for this. How could I use boolean field I am not getting it?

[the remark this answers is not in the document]
iamareebjamal: a->b->c should be reduced to a->c because we want to keep the navigation from event dashboard flat. So that no matter how many navigation a person performs, a single back click makes him come down to event dashboard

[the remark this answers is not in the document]
Prashant-Chaurasia: Yes, so for that we need to count the fragments in the stack and max 2 fragments can be in stack one is dashboard and other one is the other fragments. Thats why I have used the counter.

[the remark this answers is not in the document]
Prashant-Chaurasia: boolean field can be used if we needed to take care of one fragment only but here we need to take care of two fragments. The app is designed such that when it is created a fragment(dashboard or eventlist) is laid over the MainActivity so we need to add this fragment to the backstack and now if we select another fragment from nav bar then we need to add this one to the backstack and so total elements in backstack equals 2. So in this case we can't use boolean field.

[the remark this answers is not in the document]
iamareebjamal: Yeah so at the time of pushing, you check the size of stack to see if the limit is reached and if it is, set the boolean value. And check that boolean value here. A boolean value can be used anywhere the search space is broken into 2 by single boundary. Here there is only one if statement checking for one condition. That's a job for boolean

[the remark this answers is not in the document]
Prashant-Chaurasia: I am also doing the same thing as at the time of pushing it checks whether stack size limit is reached or not and if yes pop a fragment and then push the new one.So, where to use boolean field as such.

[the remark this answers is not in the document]
Prashant-Chaurasia: To set a boolean value we need to first check the size (using if or getbackstackentrycount method) and now for next push we check this boolean value if its true pop element and do this.But what about if we press back then we need to update this boolean field accordingly.So, they both are one or the other way similar.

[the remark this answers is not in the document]
Prashant-Chaurasia: Sir, in open-event-android app also the fragments are not added to backstack instead the fragment is replaced when pressed back button.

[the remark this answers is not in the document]
iamareebjamal: I know they are replaceable, but the way I'm describing is clean design. Representing state by what it actually means. Think of it as a component which should abstract the working of navigation and only provide control for allowable things. Soon, I'll write a more descriptive solution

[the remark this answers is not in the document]
iamareebjamal: I think Open Event Android has correct navigation flow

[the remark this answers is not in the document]
Prashant-Chaurasia: No sir, in the open event android the code is not written using backstack, instead it is replaced. Please have a look at these screenshots.
screenshot from 2018-03-22 14-58-12
screenshot from 2018-03-22 14-58-22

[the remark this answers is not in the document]
Prashant-Chaurasia: Sir, I'll do this issue please don't assign to others.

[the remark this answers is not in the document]
iamareebjamal: I know it doesn't have backstack but its navigation is alright, we don't need backstack there

app/src/main/java/org/fossasia/openevent/app/module/main/MainActivity.java
```diff
@@ -207,7 +220,28 @@ private void loadFragment(int navItemId) {
...
+
+        if(!isAtHome){
+            fragmentManager.popBackStack("root_fragment",0);
+            isAtHome=true;
+        }
+        if(atHome){
+            root_fragment_id = lastSelectedNavItemId;
+            fragment_tag = "root_fragment";
+            fragmentManager.beginTransaction().replace(R.id.fragment_container, fragment).addToBackStack(fragment_tag).commit();
+            atHome=false;
```
iamareebjamal: Both `atHome` can be extracted out of the if and made `atHome = !atHome`

app/src/main/java/org/fossasia/openevent/app/module/main/MainActivity.java
```diff
@@ -207,7 +220,28 @@ private void loadFragment(int navItemId) {
...
+        }
+        else {
+            isAtHome = false;
```
iamareebjamal: Don't understand this at all. When `atHome` is false, both the first block will run and this will run as well. Hence why not merge them. Also, one makes `atHome = true` so the `if (atHome)` is ran, then why not change the condition or just add an OR statement? The whole logic is just weird

app/src/main/java/org/fossasia/openevent/app/module/main/MainActivity.java
```diff
@@ -207,7 +220,28 @@ private void loadFragment(int navItemId) {
...
+            if((lastSelectedNavItemId==R.id.nav_dashboard)||
+                (lastSelectedNavItemId==R.id.nav_events&&root_fragment_id==lastSelectedNavItemId)){
```
iamareebjamal: Why such weird and static checks

app/src/main/java/org/fossasia/openevent/app/module/main/MainActivity.java
```diff
@@ -207,7 +220,28 @@ private void loadFragment(int navItemId) {
...
+                root_fragment_id = lastSelectedNavItemId;
```
iamareebjamal: Recommend you to read about naming variables in Java

app/src/main/java/org/fossasia/openevent/app/module/main/MainActivity.java
```diff
@@ -207,7 +220,28 @@ private void loadFragment(int navItemId) {
...
+                fragmentManager.popBackStack("root_fragment",FragmentManager.POP_BACK_STACK_INCLUSIVE);
+                fragment_tag = "root_fragment";
```
iamareebjamal: Why is fragment tag mutating so many times and `"root_fragment"` is written twice in string format when you could just use `fragment_tag`

app/src/main/java/org/fossasia/openevent/app/common/utils/ui/BackPressHandler.java
```diff
@@ -14,12 +14,17 @@
...
     public void onBackPressed(Activity activity, Runnable action) {
-        if (backPressed + BACK_PRESS_RESET_TIME > System.currentTimeMillis()) {
-            action.run();
-        } else {
-            Toast.makeText(activity, "Press again to exit", Toast.LENGTH_SHORT).show();
-        }
-        backPressed = System.currentTimeMillis();
+
+            if (backPressed + BACK_PRESS_RESET_TIME > System.currentTimeMillis()) {
+                if(activity.getClass().getSimpleName().equals("MainActivity")) {
```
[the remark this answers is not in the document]
Prashant-Chaurasia: The problem is the when you add a fragment to backstack it calls the parent activity which is here MainActivity so what the problem arose was after double click the MainActivity which was blank was visible. So, I need to check this and finish the activity after two double back press.

app/src/main/java/org/fossasia/openevent/app/module/main/MainActivity.java
```diff
@@ -207,7 +220,28 @@ private void loadFragment(int navItemId) {
                 fragment = EventDashboardFragment.newInstance(eventId);
                 break;
         }
-        fragmentManager.beginTransaction().replace(R.id.fragment_container, fragment).commit();
+
+        if(!isAtHome){
+            fragmentManager.popBackStack("root_fragment",0);
+            isAtHome=true;
+        }
+        if(atHome){
```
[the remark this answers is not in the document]
Prashant-Chaurasia: No, there are two variables one is 'atHome' and other 'isAtHome'. The variable atHome is called only when the activity is created the very first fragment(Dashboard or EventList).The logic is like this if the app is started for first time atHome block is executed. This is our main fragment to which we need to come to when pressed back. The variable 'isAtHome' describes the stack pointer whether you are on the homescreen or not.

app/src/main/java/org/fossasia/openevent/app/module/main/MainActivity.java
Prashant-Chaurasia: First 'atHome=false' is to indicate the app ran for the first time and so this block won't run after first time.Now we have 1 item in backstack. Now when we select an item from navbar it will go to else case in which 'isAtHome' indicates that the pointer is not at homefragment rather on the other fragment thus making it false.

app/src/main/java/org/fossasia/openevent/app/module/main/MainActivity.java
Prashant-Chaurasia: Suppose,we go from dashboard -> B -> dashboard. If we don't check this then what would happed is on the stack we will get dashboard->dashboard and then that would lead to improper back navigation.Same is for eventlist.

app/src/main/java/org/fossasia/openevent/app/module/main/MainActivity.java
Prashant-Chaurasia: Sorry,it was silly.It should be rootFragmentId.

app/src/main/java/org/fossasia/openevent/app/module/main/MainActivity.java
Prashant-Chaurasia: fragment_tag is changing as per the changes in the selection of nav items.

Prashant-Chaurasia:
@iamareebjamal what should I do regarding this?

iamareebjamal:
You should check why the build is failing and fix it

PR #823, opened 2018-03-22
chore: Removed Activity and Fragment instantiation tests

### LINKED ISSUE #686: Remove activity and fragment instantiation tests
**Is it a ?**

- [ ] Bug Report

- [ ] Feature Request
    
- [x] Chore

### Chore

**Description**

The activity and fragment instantiation tests are flaky and reak with every upgrade of tools. They are used to test the dependency injection and are surely first line of defence to know if the app crashes without any user interaction and the wiring of the app is correct. They also provide a lot of coverage to the project but having flaky tests is worse than having no tests, so until we have a good way to test the instantiation of activities and fragments, they should be disabled for good

sridharjajoo:
Fixes #686 

Changes: The Activity and Fragment instantiation tests were removed.

### DIFF
--- a/app/src/test/java/org/fossasia/openevent/app/robo/ui/ActivityInstanceTest.java
+++ b/app/src/test/java/org/fossasia/openevent/app/robo/ui/ActivityInstanceTest.java
@@ -14,6 +14,7 @@ import org.fossasia.openevent.app.module.main.MainActivity;
 import org.fossasia.openevent.app.module.organizer.detail.OrganizerDetailActivity;
 import org.junit.AfterClass;
 import org.junit.BeforeClass;
+import org.junit.Ignore;
 import org.junit.Test;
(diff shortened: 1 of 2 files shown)


iamareebjamal:
Don't remove the classes, just comment out the annotations

sridharjajoo:
I have removed the unused imports as well

iamareebjamal:
@sridharjajoo Please fix he build

sridharjajoo:
@iamareebjamal As the annotations have been commented out , the build is failing.

iamareebjamal:
You have to test the build locally that it is working before sending the PR. Not all annotations need to be disabled. Check which ones need to stay on

sridharjajoo:
@iamareebjamal Instead of commenting out , can I use the @Ignore notation for the classes?

PR #829, opened 2018-03-23
feat: Error feedback from JSON response

### LINKED ISSUE #593: Error handling of JSON response
**Is it a ?**

- [ ] Bug Report

- [x] Feature Request
    
- [ ] Chore

**Feature Description**

The error response given by server as JSON String is currently not being deserialized which could instead give better error description for the request. Currently the error string is provided to the view by ErrorUtils which gives the response on the basis of HTTP Error code but better detailed error response is possible by deserializing the server response and passing the string in "detail" tag of the JSON response as the error response to the view.
   
Yes

Masquerade0097:
Fixes #593 

Changes: 
1. Provide Error Feedback by parsing Response JSON.
2. Added Unit Tests for ErrorUtils.
3. Load JSONs from Resources in JUnit tests.

Description: 
Error feedback is provided to user by parsing the JSON error response by server. The response is comprised of the JSON String under `detail` tag concatenated with the field pointed by `pointer` tag.
 
Error JSON : `{"errors": [{"status": "422", "source": {"pointer": "/data/attributes/licence"}, "detail": "Missing data for required field.", "title": "Validation error"}], "jsonapi": {"version": "1.0"}}`

Error feedback provided by app :  `Missing data for required field: license`

Screenshots for the change: 
ervd1


app/src/main/java/org/fossasia/openevent/app/common/utils/core/ErrorUtils.java
```diff
@@ -1,39 +1,49 @@
...
-    public static final int FORBIDDEN = 403;
-    public static final int NOT_FOUND = 404;
-    public static final int METHOD_NOT_ALLOWED = 405;
-    public static final int REQUEST_TIMEOUT = 408;
-
     private ErrorUtils() {
         // Never Called
     }
 
     public static String getMessage(Throwable throwable) {
```
iamareebjamal: Nice job. Can you please add tests for these as well

Masquerade0097: Apologies for furthur delay but I'm quite busy with my mid sems and proposal, will update it soon.

iamareebjamal:
@Masquerade0097 If you're free, please update this. Also, you have removed previous errors. Please add them back as well

Masquerade0097:
@iamareebjamal By previous errors do you mean that the app should show the previous error messages based on the error codes which are hardcoded and for the others it should parse the error JSON ?

iamareebjamal:
Yes. A better alternative would be to parse the throwable and set the response in a custom exception and throw that instead. But for now, this will do

Masquerade0097:
I'm unable to test getErrorDetails using Mockito because I'm allocating `JSONObject` inside `getErrorDetails()` that's why I'm using Robolectric to do so.

iamareebjamal:
JSONObject works in tests as well. See JWTUtils tets

Masquerade0097:
@iamareebjamal When I'm running with JUnit the test no 3,4,5,6 fails. The only solution I found out was to use 'org.json:json:20180130` for tests so that Android SDK is not used. The build.gradle has `testImplementation 'org.json:json:20180130`, also I've added `Configurations.resolutionStrategy force 'org.json:json:20180130'` but still I'm getting null string in asserts from the tests no 3,4,5,6 and they fails.
On Using Robolectric all these tests complete successfully.

app/src/test/java/org/fossasia/openevent/app/unit/utils/ErrorUtilsTest.java
```diff
@@ -0,0 +1,108 @@
...
+import static org.junit.Assert.assertNull;
+
+//@RunWith(RobolectricTestRunner.class)
+//@Config(sdk = Config.OLDEST_SDK, application = Application.class)
+@RunWith(JUnit4.class)
+public class ErrorUtilsTest {
+
+    public static final String CONTENT1 = "{\"errors\": [{\"status\": \"422\", \"source\": {\"pointer\": \"/data/attributes/licence\"}, \"detail\": \"Missing data for required field.\", \"title\": \"Validation error\"}], \"jsonapi\": {\"version\": \"1.0\"}}";
+    public static final String CONTENT2 = "{\"errors\": [{\"status\": \"422\", \"source\": {\"pointer\": \"\"}, \"detail\": \"Missing data for required field.\", \"title\": \"Validation error\"}], \"jsonapi\": {\"version\": \"1.0\"}}";
+    public static final String CONTENT3 = "{\"errors\": [{\"status\": \"422\", \"source\": {}, \"detail\": \"Missing data for required field.\", \"title\": \"Validation error\"}], \"jsonapi\": {\"version\": \"1.0\"}}";
```
iamareebjamal: Add these as JSONs in resources directory and load from there

app/src/main/java/org/fossasia/openevent/app/utils/ErrorUtils.java
```diff
@@ -19,7 +28,7 @@ public static String getMessage(Throwable throwable) {
         if (throwable instanceof HttpException) {
             switch (((HttpException) throwable).code()) {
                 case BAD_REQUEST:
-                    return "Something went wrong! Please check any empty field of a form.";
+                    return "Something went wrong! Please check any empty field if a form.";
```
iamareebjamal: of

app/src/test/resources/raw/content1.json
```diff
@@ -0,0 +1 @@
+{"errors": [{"status": "422", "source": {"pointer": "/data/attributes/licence"}, "detail": "Missing data for required field.", "title": "Validation error"}], "jsonapi": {"version": "1.0"}}
```
iamareebjamal: Format the JSONs properly

Masquerade0097:
@iamareebjamal Changes done. Please review.

app/src/test/java/org/fossasia/openevent/app/unit/utils/ErrorUtilsTest.java
```diff
@@ -0,0 +1,139 @@
...
+        String str = String.valueOf(ErrorUtils.getMessage(httpException));
+
+        assertEquals("Something went wrong! Please check any empty field of a form.", str);
+    }
+
+    @Test
+    public void shouldReturnOtherThrowableErrorMessageSuccessfully() {
+        String str = String.valueOf(ErrorUtils.getMessage(ioException));
+
+        assertEquals("null", str);
```
iamareebjamal: Why is return string "null"

app/src/test/java/org/fossasia/openevent/app/unit/utils/ErrorUtilsTest.java
```diff
@@ -0,0 +1,139 @@
...
+        assertNull(ErrorUtils.getPointedField(pointer2));
+        assertNull(ErrorUtils.getPointedField(pointer3));
+    }
+
+    @Test
+    public void shouldReturnErrorDetailsWithPointedFieldSuccessfully() {
+        String str = String.valueOf(ErrorUtils.getErrorDetails(ERROR_RESPONSE_4.errorBody()));
+
+        assertEquals("Missing data for required field: licence", str);
+        assertNotEquals("Some random string", str);
```
iamareebjamal: Remove this line from tests

app/src/test/java/org/fossasia/openevent/app/unit/utils/ErrorUtilsTest.java
Masquerade0097: The `IOException` thrown in the test has empty message body so `throwable.getMessage()` returns `"null"` but in the application this will most likely not be the case because the system won't throw an empty IOException. I just wanted to highlight the case that's why wrote this test.

PR #840, opened 2018-03-25
feat: Allow deleting Copyright of the Event

### LINKED ISSUE #738: feat: Allow deleting the copyright
**Is it a ?**

- [ ] Bug Report

- [x] Feature Request
    
- [ ] Chore
    
### Feature Request

**Feature Description**

Right now the application supports creating Copyright for an Event. It should also support deleting the copyright.
   
Yes

**Parent issue**

#619 

To be done after #737

srv-twry:
Fixes #738 

Changes: Added an option to delete the Copyright of the Event if it exists

Screenshots for the change: 
videotogif_2018 03 25_16 59 32

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/app/module/event/about/AboutEventPresenter.java
+++ b/app/src/main/java/org/fossasia/openevent/app/module/event/about/AboutEventPresenter.java
@@ -1,5 +1,7 @@
 package org.fossasia.openevent.app.module.event.about;
 
+import android.support.annotation.Nullable;
+
 import com.raizlabs.android.dbflow.structure.BaseModel;
 
 import org.fossasia.openevent.app.common.app.lifecycle.presenter.BaseDetailPresenter;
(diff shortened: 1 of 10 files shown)


app/src/main/java/org/fossasia/openevent/app/module/event/about/AboutEventPresenter.java
```diff
@@ -90,6 +94,19 @@ public Copyright getCopyright() {
...
 
+    @Override
+    public void deleteCopyright() {
+        copyrightRepository
+            .deleteCopyright(copyright.getId())
+            .compose(disposeCompletable(getDisposable()))
+            .compose(progressiveErroneousCompletable(getView()))
+            .subscribe(() -> {
+                getView().showCopyrightDeleted("Copyright Deleted");
+                copyright = null;
```
srv-twry: Need your help here @iamareebjamal 
I am having to add this line due to which i am getting a PMD voilation due to which the build fails.
I have to add this line because once the user deletes the copyright, the API return HTTP 404 which is sent to the logger as an error. It means in the ```loadCopyright``` method, the ```onComplete``` won't run which would have assigned the ```null``` value to the copyright variable, instead ```onError``` would be triggered here 

```subscribe(loadedCopyright -> this.copyright = loadedCopyright, Logger::logError);```

iamareebjamal: You can suppress the PMD warning

PR #859, opened 2018-03-30
fix: Remove static Copyright field from UpdateCopyrightFragment

### LINKED ISSUE #852: fix: Remove static copyright field from the Update Copyright Fragment
**Is it a ?**

- [ ] Bug Report

- [ ] Feature Request
    
- [x] Chore
    
### Chore

**Description**

There is a static copyright field inside the Update Copyright Fragment. It needs to be removed. 
Moreover the following lines are unnecessary
``` Java
final Context contextThemeWrapper = new ContextThemeWrapper(getActivity(), R.style.AppTheme);
LayoutInflater localInflater = inflater.cloneInContext(contextThemeWrapper);
```

Yes

srv-twry:
Fixes #852 

Changes: Removed static Copyright field from the UpdateCopyrightFragment
Copyright fetched from the database.

Screenshots for the change: 
videotogif_2018 03 30_13 00 47


srv-twry:
@iamareebjamal One query.
Can the application fail to load the Copyright from the db even if it exists in the db ? I mean should i handle this case also. If it fails to load due to some error, the user will be presented with a blank form. May be i can show an error that "Copyright failed to load" and close the fragment.
What do you suggest ?

iamareebjamal:
No, if it exists in db, only then we are going to show edit form so there is no need for that extra check

PR #865, opened 2018-03-31
feat: Implement spinner for 'Payment Country', choose default, and au…

### LINKED ISSUE #732: Implement spinner for 'Payment Country', choose default, and auto-select 'Payment Currency' using it.
**Is it a ?**

- [x] Bug Report

- [x] Feature Request
    
- [ ] Chore
    
### Feature Request

**Feature Description**

Since we can automatically detect organizer's timezone and currency based on locale, we probably also can auto-fill payment country. Maybe we should make it a spinner, just like timezone and currency instead of a EditText field.

Also, in the present state, country name is not validated. Therefore, any random name can be filled. This qualifies it as a bug too.

Yes

mishuvs:
…to-select 'Payment Currency' using it

Fixes #732 

Changes: 
1. Add spinner for Payment Country
2. Replace code to set default 'Currency Code' with code to set default Payment Country based on Locale
3. Set default Currency Code using Payment Country selection
4. Rename 'CurrencyUtil.java' to 'CountryUtil.java'
5. Reduce Cyclomatic Complexity for CreateEventFragment.java

Screenshots of the change: 

country

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/app/core/event/create/CreateEventFragment.java
+++ b/app/src/main/java/org/fossasia/openevent/app/core/event/create/CreateEventFragment.java
@@ -15,6 +15,7 @@ import android.text.TextWatcher;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
+import android.widget.AdapterView;
 import android.widget.ArrayAdapter;
 import android.widget.Toast;
 
@@ -52,7 +53,8 @@ public class CreateEventFragment extends BaseBottomSheetFragment<CreateEventPres
 
     private EventCreateLayoutBinding binding;
     private Validator validator;
-    private ArrayAdapter<CharSequence> adapter;
+    private ArrayAdapter<CharSequence> currencyAdapter;
+    private ArrayAdapter<CharSequence> paymentCountryAdapter;
 
     private static final int PLACE_PICKER_REQUEST = 1;
     private final GoogleApiAvailability googleApiAvailabilityInstance = GoogleApiAvailability.getInstance();
@@ -82,50 +84,32 @@ public class CreateEventFragment extends BaseBottomSheetFragment<CreateEventPres
                 getPresenter().createEvent();
         });
 
-        //check if there's an google places API key
-        try {
-            ApplicationInfo ai = getContext().getPackageManager().getApplicationInfo(getContext().getPackageName(), PackageManager.GET_META_DATA);
-            Bundle bundle = ai.metaData;
-            String placesApiKey = bundle.getString("com.google.android.geo.API_KEY");
-            if ("YOUR_API_KEY".equals(placesApiKey)) {
-                Timber.d("Add Google Places API key in AndroidManifest.xml file to use Place Picker.");
-                binding.form.buttonPlacePicker.setVisibility(View.GONE);
-                binding.form.layoutLatitude.setVisibility(View.VISIBLE);
-                binding.form.layoutLongitude.setVisibility(View.VISIBLE);
-                showLocationLayouts();
-            }
-        } catch (PackageManager.NameNotFoundException e) {
-            Timber.e(e, "Package name not found");
-        }
+        setupSpinners();
 
-        binding.form.buttonPlacePicker.setOnClickListener(view -> {
-            int errorCode = googleApiAvailabilityInstance.isGooglePlayServicesAvailable(getContext());
-            if (errorCode == ConnectionResult.SUCCESS) {
-                //SUCCESS
-                PlacePicker.IntentBuilder builder = new PlacePicker.IntentBuilder();
-                try {
-                    startActivityForResult(builder.build(getActivity()), PLACE_PICKER_REQUEST);
-                } catch (GooglePlayServicesRepairableException e) {
-                    Timber.d(e, "GooglePlayServicesRepairable");
-                } catch (GooglePlayServicesNotAvailableException e) {
-                    Timber.d("GooglePlayServices NotAvailable => Updating or Unauthentic");
-                }
-            } else if (googleApiAvailabilityInstance.isUserResolvableError(errorCode)) {
-                //SERVICE_MISSING, SERVICE_VERSION_UPDATE_REQUIRED, SERVICE_DISABLED
-                googleApiAvailabilityInstance.getErrorDialog(
-                    getActivity(), errorCode, PLACE_PICKER_REQUEST, (dialog) -> {
-                        showLocationLayouts();
-                    });
-            } else {
-                //SERVICE_UPDATING, SERVICE_INVALID - can't use place picker - must enter manually
-                showLocationLayouts();
-            }
-        });
-
-        adapter = new ArrayAdapter<>(getActivity(), android.R.layout.simple_spinner_item);
-        adapter.setDropDownViewResource(android.R.layout.simple_spinner_dropdown_item);
+        setupPlacePicker();
 
         return binding.getRoot();
+    }
+
+    private void setupSpinners() {
+        currencyAdapter = new ArrayAdapter<>(getActivity(), android.R.layout.simple_spinner_item);
+        currencyAdapter.setDropDownViewResource(android.R.layout.simple_spinner_dropdown_item);
+        paymentCountryAdapter = new ArrayAdapter<>(getActivity(), android.R.layout.simple_spinner_item);
+        paymentCountryAdapter.setDropDownViewResource(android.R.layout.simple_spinner_dropdown_item);
+
+        binding.form.paymentCountrySpinner.setOnItemSelectedListener(new AdapterView.OnItemSelectedListener() {
+
+            @Override
+            public void onItemSelected(AdapterView<?> adapterView, View view, int i, long l) {
+                getPresenter().onPaymentCountrySelected(adapterView.getItemAtPosition(i).toString());
+            }
+
+            @Override
+            public void onNothingSelected(AdapterView<?> adapterView) {
+                //do nothing
+            }
+
+        });
     }
 
     @Override
@@ -174,9 +158,20 @@ public class CreateEventFragment extends BaseBottomSheetFragment<CreateEventPres
     }
 
     @Override
+    public void setPaymentCurrency(int index) {
+        binding.form.currencySpinner.setSelection(index);
+    }
+
+    @Override
+    public void attachCountryList(List<String> countryList) {
+        paymentCountryAdapter.addAll(countryList);
+        binding.form.paymentCountrySpinner.setAdapter(paymentCountryAdapter);
+    }
+
+    @Override
     public void attachCurrencyCodesList(List<String> currencyCodesList) {
-        adapter.addAll(currencyCodesList);
-        binding.form.currencySpinner.setAdapter(adapter);
+        currencyAdapter.addAll(currencyCodesList);
+        binding.form.currencySpinner.setAdapter(currencyAdapter);
     }
 
     @Override
@@ -215,8 +210,50 @@ public class CreateEventFragment extends BaseBottomSheetFragment<CreateEventPres
     }
 
     @Override
-    public void setDefaultCurrency(int index) {
-        binding.form.currencySpinner.setSelection(index);
+    public void setDefaultCountry(int index) {
+        binding.form.paymentCountrySpinner.setSelection(index);
+    }
+
+    private void setupPlacePicker() {
+        //check if there's an google places API key
+        try {
+            ApplicationInfo ai = getContext().getPackageManager().getApplicationInfo(getContext().getPackageName(), PackageManager.GET_META_DATA);
+            Bundle bundle = ai.metaData;
+            String placesApiKey = bundle.getString("com.google.android.geo.API_KEY");
+            if ("YOUR_API_KEY".equals(placesApiKey)) {
+                Timber.d("Add Google Places API key in AndroidManifest.xml file to use Place Picker.");
+                binding.form.buttonPlacePicker.setVisibility(View.GONE);
+                binding.form.layoutLatitude.setVisibility(View.VISIBLE);
+                binding.form.layoutLongitude.setVisibility(View.VISIBLE);
+                showLocationLayouts();
+            }
+        } catch (PackageManager.NameNotFoundException e) {
+            Timber.e(e, "Package name not found");
+        }
+
+        binding.form.buttonPlacePicker.setOnClickListener(view -> {
+            int errorCode = googleApiAvailabilityInstance.isGooglePlayServicesAvailable(getContext());
+            if (errorCode == ConnectionResult.SUCCESS) {
+                //SUCCESS
+                PlacePicker.IntentBuilder builder = new PlacePicker.IntentBuilder();
+                try {
+                    startActivityForResult(builder.build(getActivity()), PLACE_PICKER_REQUEST);
+                } catch (GooglePlayServicesRepairableException e) {
+                    Timber.d(e, "GooglePlayServicesRepairable");
+                } catch (GooglePlayServicesNotAvailableException e) {
+                    Timber.d("GooglePlayServices NotAvailable => Updating or Unauthentic");
+                }
+            } else if (googleApiAvailabilityInstance.isUserResolvableError(errorCode)) {
+                //SERVICE_MISSING, SERVICE_VERSION_UPDATE_REQUIRED, SERVICE_DISABLED
+                googleApiAvailabilityInstance.getErrorDialog(
+                    getActivity(), errorCode, PLACE_PICKER_REQUEST, (dialog) -> {
+                        showLocationLayouts();
+                    });
+            } else {
+                //SERVICE_UPDATING, SERVICE_INVALID - can't use place picker - must enter manually
+                showLocationLayouts();
+            }
+        });
     }
 
     @Override
--- a/app/src/main/java/org/fossasia/openevent/app/core/event/create/CreateEventPresenter.java
+++ b/app/src/main/java/org/fossasia/openevent/app/core/event/create/CreateEventPresenter.java
@@ -11,9 +11,10 @@ import org.threeten.bp.LocalDateTime;
 import org.threeten.bp.ZonedDateTime;
 import org.threeten.bp.format.DateTimeParseException;
 
-import java.util.Currency;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
+import java.util.Map;
 import java.util.TimeZone;
 
 import javax.inject.Inject;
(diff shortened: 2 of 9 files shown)


app/src/main/java/org/fossasia/openevent/app/core/event/create/CreateEventFragment.java
```diff
@@ -52,7 +55,7 @@
 
     private EventCreateLayoutBinding binding;
     private Validator validator;
-    private ArrayAdapter<CharSequence> adapter;
+    private ArrayAdapter<CharSequence> currencyAdapter, paymentCountryAdapter;
```
iamareebjamal: declaration in separate line

app/src/main/java/org/fossasia/openevent/app/core/event/create/CreateEventFragment.java
```diff
@@ -82,50 +85,36 @@ public View onCreateView(LayoutInflater inflater, @Nullable ViewGroup container,
...
+    private void setupSpinners() {
+        currencyAdapter = new ArrayAdapter<>(getActivity(), android.R.layout.simple_spinner_item);
+        currencyAdapter.setDropDownViewResource(android.R.layout.simple_spinner_dropdown_item);
+        paymentCountryAdapter = new ArrayAdapter<>(getActivity(), android.R.layout.simple_spinner_item);
+        paymentCountryAdapter.setDropDownViewResource(android.R.layout.simple_spinner_dropdown_item);
+        binding.form.paymentCountrySpinner.setOnItemSelectedListener(new AdapterView.OnItemSelectedListener() {
+
+            @Override
+            public void onItemSelected(AdapterView<?> adapterView, View view, int i, long l) {
+                getPresenter().getEvent().paymentCountry = adapterView.getItemAtPosition(i).toString();
```
iamareebjamal: View should not handle any logic

app/src/main/java/org/fossasia/openevent/app/core/event/create/CreateEventFragment.java
```diff
@@ -82,50 +85,36 @@ public View onCreateView(LayoutInflater inflater, @Nullable ViewGroup container,
...
+                Map<String, String> countryCurrencyMap = CountryUtils.getCountryCurrencyMap();
```
iamareebjamal: Getting currency map each time an item is selected

app/src/main/java/org/fossasia/openevent/app/utils/CountryUtils.java
```diff
@@ -1,24 +1,25 @@
...
 
 import io.reactivex.Single;
 import io.reactivex.schedulers.Schedulers;
+import timber.log.Timber;
 
 /**
  * Source: https://stackoverflow.com/a/15343675/3309666
  */
-public final class CurrencyUtils {
+public final class CountryUtils {
```
iamareebjamal: Revert

app/src/main/java/org/fossasia/openevent/app/utils/CountryUtils.java
```diff
@@ -35,7 +36,7 @@ private CurrencyUtils() {
                     Currency currency = Currency.getInstance(locale);
                     currencyLocaleMap.put(currency, locale);
                 } catch (Exception e) {
-                    // No action
+                    Timber.d(e);
```
iamareebjamal: No action

app/src/main/java/org/fossasia/openevent/app/utils/CountryUtils.java
```diff
@@ -57,33 +58,47 @@ private CurrencyUtils() {
...
-        priorCurrencyCodes.add(Currency.getInstance(Locale.CANADA).toString());
-        priorCurrencyCodes.add(Currency.getInstance(Locale.CHINA).toString());
-        priorCurrencyCodes.add(Currency.getInstance(Locale.KOREA).toString());
-
-        for (Locale locale : Locale.getAvailableLocales()) {
+    public static Map<String, String> getCountryCurrencyMap() {
+        Locale[] locales = Locale.getAvailableLocales();
+        Map<String, String> countryCurrencyMap = new TreeMap<>();
+        String country;
+        Currency currency;
```
iamareebjamal: Why are county and currency defined outside the loop

app/src/main/java/org/fossasia/openevent/app/utils/CountryUtils.java
```diff
@@ -57,33 +58,47 @@ private CurrencyUtils() {
...
-        priorCurrencyCodes.add(Currency.getInstance(Locale.UK).toString());
-        priorCurrencyCodes.add(Currency.getInstance(Locale.GERMANY).toString());
-        priorCurrencyCodes.add(Currency.getInstance(Locale.JAPAN).toString());
-        priorCurrencyCodes.add(Currency.getInstance(Locale.CANADA).toString());
-        priorCurrencyCodes.add(Currency.getInstance(Locale.CHINA).toString());
-        priorCurrencyCodes.add(Currency.getInstance(Locale.KOREA).toString());
-
-        for (Locale locale : Locale.getAvailableLocales()) {
+    public static Map<String, String> getCountryCurrencyMap() {
+        Locale[] locales = Locale.getAvailableLocales();
```
iamareebjamal: No need to store in a local variable

app/src/main/java/org/fossasia/openevent/app/core/event/create/CreateEventFragment.java
mishuvs: 1. I am unable to bind `paymentCountry` in XML.
2. Invoking a `presenter` method using lambda expression in XML seems to not work. 
3. Calling a presenter method from Listener in Fragment will require one more method in Fragment that the Presenter can call, or else Presenter will require accessing Binding variable.
4. A lengthier solution is to use inverse data binding.

Please suggest how to handle this.

app/src/main/java/org/fossasia/openevent/app/utils/CountryUtils.java
iamareebjamal: Only the reference is outside the loop. You are not reusing variables to be declared outside. New allocation is still done in the loop

```Java
SharedObject so = new SharedObject();
for (Item i : items {
    so.doSomething();
    // stuff
}
```

This is better

```Java
SharedObject so;
for (Item i : items {
    so = new SharedObject()
    so.doSomething();
    // stuff
}
```
This is not

Also, when there is no object that is loop invariant which is the case here, then

```Java
String name;
String title;
for (Item i : items) {
    name = item.getName();
    title = item.getTitle();
}
```

is worse than

```Java
for (Item i : items) {
    String name = item.getName();
    String title = item.getTitle();
}
```

because it
- has no performance benefit
- increases the scope of variables outside the loop
- makes the variables nullable

Also, it should be noted that simplicity, readability and maintainability are favoured far more than performance. Optimisation is the last thing to be considered after profiling and checking of the bottleneck is done. Adding to this, however, it is important to follow good practices related to performance as you mentioned, but only knowing when they are applied

app/src/main/java/org/fossasia/openevent/app/core/event/create/CreateEventFragment.java
iamareebjamal: You shouldn't do it in XML at all. The view should just tell presenter that this event has happened and presenter should handle the rest

app/src/main/java/org/fossasia/openevent/app/utils/CountryUtils.java
mishuvs: Thanks for sharing your knowledge and for the very detailed answer. I sure am learning a lot from you. I'll make the change and push.

app/src/main/java/org/fossasia/openevent/app/core/event/create/CreateEventFragment.java
mishuvs: @iamareebjamal I believe you meant approach no. 3. 
I have made changes. Please review.

mishuvs:
@iamareebjamal Please review.

app/src/main/java/org/fossasia/openevent/app/core/event/create/CreateEventPresenter.java
```diff
@@ -1,5 +1,7 @@
 package org.fossasia.openevent.app.core.event.create;
 
+import android.content.res.Resources;
```
iamareebjamal: No `android.*` import is allowed in a presenter

mishuvs: I'll need context to get the default locale. View can't handle logic. 
Do I need to inject context?

app/src/main/java/org/fossasia/openevent/app/core/event/create/CreateEventFragment.java
```diff
@@ -173,10 +158,21 @@ public void afterTextChanged(Editable editable) {
         });
     }
 
+    @Override
+    public void setPaymentCurrency(String paymentCurrency) {
+        binding.form.currencySpinner.setSelection(CurrencyUtils.getCurrencyCodesList().indexOf(paymentCurrency));
```
iamareebjamal: Simply pass the index from presenter.

app/src/main/java/org/fossasia/openevent/app/core/event/create/CreateEventPresenter.java
```diff
@@ -38,17 +37,18 @@ public CreateEventPresenter(IEventRepository eventRepository) {
...
-        //set default timezone
         List<String> timeZoneList = getView().getTimeZoneList();
         getView().setDefaultTimeZone(
             timeZoneList.indexOf(TimeZone.getDefault().getID())
         );
 
-        List<String> currencyList = CurrencyUtils.getCurrencyCodesList();
-        getView().setDefaultCurrency(
-            currencyList.indexOf(Currency.getInstance(Locale.getDefault()).getCurrencyCode())
+        List<String> countryList = CurrencyUtils.getCountryList();
```
iamareebjamal: Cache country list rather than calling again and again

app/src/main/java/org/fossasia/openevent/app/core/event/create/CreateEventPresenter.java
```diff
@@ -109,4 +109,15 @@ public String getSearchableLocationName(String address) {
...
     }
+
+    /**
+     * auto-selects paymentCurrency when paymentCountry is selected.
+     * @param paymentCountry chosen payment country
+     */
+    void onPaymentCountrySelected(String paymentCountry) {
+        event.paymentCountry = paymentCountry;
+        String paymentCurrency = CurrencyUtils.getCountryCurrencyMap().get(paymentCountry);
+        event.paymentCurrency = paymentCurrency;
```
iamareebjamal: Always use setters

app/src/main/java/org/fossasia/openevent/app/core/event/create/CreateEventPresenter.java
```diff
@@ -109,4 +109,15 @@ public String getSearchableLocationName(String address) {
...
         } else return primary;
     }
+
+    /**
+     * auto-selects paymentCurrency when paymentCountry is selected.
+     * @param paymentCountry chosen payment country
+     */
+    void onPaymentCountrySelected(String paymentCountry) {
+        event.paymentCountry = paymentCountry;
+        String paymentCurrency = CurrencyUtils.getCountryCurrencyMap().get(paymentCountry);
```
iamareebjamal: Cache country map

app/src/main/java/org/fossasia/openevent/app/utils/CurrencyUtils.java
```diff
@@ -57,33 +58,43 @@ private CurrencyUtils() {
...
-        List<String> currencyCodes = new ArrayList<>();
-        List<String> priorCurrencyCodes = new ArrayList<>();
-
-        priorCurrencyCodes.add(Currency.getInstance(Locale.US).toString());
-        priorCurrencyCodes.add(Currency.getInstance(Locale.UK).toString());
-        priorCurrencyCodes.add(Currency.getInstance(Locale.GERMANY).toString());
-        priorCurrencyCodes.add(Currency.getInstance(Locale.JAPAN).toString());
-        priorCurrencyCodes.add(Currency.getInstance(Locale.CANADA).toString());
-        priorCurrencyCodes.add(Currency.getInstance(Locale.CHINA).toString());
-        priorCurrencyCodes.add(Currency.getInstance(Locale.KOREA).toString());
```
iamareebjamal: These currencies were given priority

app/src/main/java/org/fossasia/openevent/app/utils/CurrencyUtils.java
```diff
@@ -57,33 +58,43 @@ private CurrencyUtils() {
...
+        ArrayList<String> countries = new ArrayList<>();
+
+        for (Map.Entry<String, String> entry : countryCurrencyMap.entrySet()) {
+            countries.add(entry.getKey());
+        }
 
-        Collections.sort(currencyCodes);
-        priorCurrencyCodes.addAll(currencyCodes);
-        return priorCurrencyCodes;
+        return countries;
```
iamareebjamal: Why is there a need for 3 functions

app/src/main/java/org/fossasia/openevent/app/utils/CurrencyUtils.java
mishuvs: Priority was given so that the users could select these countries without scrolling much. In the new state, `currency` will automatically be selected in the spinner. So there's no question of scrolling as any item in the middle will be selected as per locale.
From there, if the user wants to change to any one of the priority countries, he will scroll alphabetically and not to the top.

In the loop:
```
for (Locale locale : Locale.getAvailableLocales()) {
            try {
                Currency currency = Currency.getInstance(locale);
                if (currency != null && !currencyCodes.contains(currency.toString())) {
                    currencyCodes.add(currency.toString());
                }
            } catch (Exception e) {
                // No action
            }
        }
```
`priorCurrencyCodes` were being added to the list without checking if they exist:
`priorCurrencyCodes.addAll(currencyCodes);`
I think this is repetitive, and creates a bit of confusion and bad UX.

app/src/main/java/org/fossasia/openevent/app/core/event/create/CreateEventPresenter.java
mishuvs: It lives in `Presenter#start` method, which is called only once. How is this being used again and again?

app/src/main/java/org/fossasia/openevent/app/utils/CurrencyUtils.java
iamareebjamal: In almost all pickers of currency, you see common ones on the top. Default selected or not. It is very common for someone to live in India and choose USD as default currency

app/src/main/java/org/fossasia/openevent/app/core/event/create/CreateEventPresenter.java
iamareebjamal: It's used twice in `start`

app/src/main/java/org/fossasia/openevent/app/utils/CurrencyUtils.java
mishuvs: We need sorted `List<String>`s for both spinners. We can create them using `List<Value> list = new ArrayList<Value>(map.values());` or `map.entrySet()`

But should be noted that the order of values returned by those methods is not defined and for HashMap and similar un-sorted Map implementations it will be effectively random.
So every time we need `currencyCodesList` or `CountriesList`, we'll have to sort.

I do agree that the methods need modification to simply use `Collections.sort`. Please tell if there's any better solution.

iamareebjamal: The thing is the calculation of these lists or maps is considered an expensive operation. This function was only called once previously. Now it's called many times and in form of three functions.

We can cache the result inside the utils class itself and return precomputed result instead. In order to do so, make the methods non static and store the precomputed result in private members. And add `@Inject` annotation on the constructor of the class after making it public

LinkedHashMap is sorted by the order of entry

mishuvs: We iterate using the values from `#getAvailableLocales`, in which some countries don't follow alphabetical order. 
TreeMap, on the other hand, is sorted by keys.

app/src/main/java/org/fossasia/openevent/app/utils/CurrencyUtils.java
mishuvs: I agree that since I live in India, 'INR' will be selected even when 'USD' is needed. But then, to select 'USD' **the most natural action would be to scroll downwards to reach the letter 'U'**. I wouldn't even know/remember that you had placed USD at the top for my convenience.

Google Translate also handles country and language (language analogous to currency in our case) selection similarly:

**Spanish and English was selected by default**

gtranslate

iamareebjamal: The list is huge and some of the countries currency are used very often. If we add them to the top and don't remove them from bottom, I guess both problems will be solved

mishuvs:
@iamareebjamal Please review.
And I wanted to ask if we should squash our commits at the end, or after every incremental commit.
Also, apologies for the delay. I was caught up in college's internal evals and submissions.

app/src/main/java/org/fossasia/openevent/app/utils/CurrencyUtils.java
```diff
@@ -1,24 +1,30 @@
...
  * Source: https://stackoverflow.com/a/15343675/3309666
  */
 public final class CurrencyUtils {
 
     private static SortedMap<Currency, Locale> currencyLocaleMap;
+    private static Map<String, String> countryCurrencyMap;
 
-    private CurrencyUtils() {
+    @Inject
+    public CurrencyUtils() {
```
iamareebjamal: This is a static utils class, if it is getting injected, make sure each static instance and method is made non static

app/src/main/java/org/fossasia/openevent/app/utils/CurrencyUtils.java
```diff
@@ -70,19 +91,8 @@ private CurrencyUtils() {
...
-            try {
-                Currency currency = Currency.getInstance(locale);
-                if (currency != null && !currencyCodes.contains(currency.toString())) {
-                    currencyCodes.add(currency.toString());
-                }
-            } catch (Exception e) {
-                // No action
-            }
-        }
+        priorCurrencyCodes.addAll(new ArrayList<>(new TreeSet<>(countryCurrencyMap.values())));
```
iamareebjamal: NullPointerException

app/src/main/java/org/fossasia/openevent/app/utils/CurrencyUtils.java
```diff
@@ -57,9 +63,24 @@ private CurrencyUtils() {
             .subscribeOn(Schedulers.computation());
     }
 
-    @SuppressWarnings({ "PMD.AvoidCatchingGenericException", "PMD.EmptyCatchBlock" })
-    public static List<String> getCurrencyCodesList() {
-        List<String> currencyCodes = new ArrayList<>();
+    @SuppressWarnings({ "PMD.UseConcurrentHashMap"})
+    public Map<String, String> getCountryCurrencyMap() {
+        countryCurrencyMap = new TreeMap<>();
```
iamareebjamal: Creating country map again and again even if member may not be null

iamareebjamal:
@mishuvs It's better to squash as soon as possible because it makes rebasing with development easier

PR #873, opened 2018-04-05
Added databasechange listener to eventslist

### LINKED ISSUE #690: Event List not updating automatically after creation
**Is it a ?**

- [x] Bug Report

- [ ] Feature Request
    
- [ ] Chore
    
### Bug Report    

**Actual Behaviour**

After an event is created, the list is not updated automatically

**Expected Behaviour**

- [ ] Once the event is created, it's response is the event itself and hence should be saved in the repository on receiving.
- [ ] Then, a change listener should be created in the event list presenter, so that any change in database is reflected in the list

sridharjajoo:
Fixes #690 
Changes:

- [x] A database change listener was added.
- [x] Events should show up automatically after creation of new event.

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/app/core/event/create/CreateEventFragment.java
+++ b/app/src/main/java/org/fossasia/openevent/app/core/event/create/CreateEventFragment.java
@@ -7,7 +7,6 @@ import android.databinding.DataBindingUtil;
 import android.os.Bundle;
 import android.support.annotation.Nullable;
 import android.support.design.widget.TextInputLayout;
-import android.support.v7.app.ActionBar;
 import android.support.v7.app.AppCompatActivity;
 import android.text.Editable;
 import android.text.TextUtils;
@@ -68,17 +67,6 @@ public class CreateEventFragment extends BaseBottomSheetFragment<CreateEventPres
         binding = DataBindingUtil.inflate(inflater, R.layout.event_create_layout, container, false);
         validator = new Validator(binding.form);
 
-        AppCompatActivity activity = ((AppCompatActivity) getActivity());
-        activity.setSupportActionBar(binding.toolbar);
-
-        ActionBar actionBar = activity.getSupportActionBar();
-        if (actionBar != null) {
-            actionBar.setHomeButtonEnabled(true);
-            actionBar.setDisplayHomeAsUpEnabled(true);
-        }
-
-        setHasOptionsMenu(true);
-
         binding.submit.setOnClickListener(view -> {
             if (validator.validate())
                 getPresenter().createEvent();
@@ -118,6 +106,7 @@ public class CreateEventFragment extends BaseBottomSheetFragment<CreateEventPres
         getPresenter().attach(this);
         binding.setEvent(getPresenter().getEvent());
         getPresenter().start();
+        ((AppCompatActivity)getActivity()).getSupportActionBar().hide();
 
         validate(binding.form.ticketUrlLayout, ValidateUtils::validateUrl, getResources().getString(R.string.url_validation_error));
         validate(binding.form.logoUrlLayout, ValidateUtils::validateUrl, getResources().getString(R.string.url_validation_error));
@@ -196,7 +185,8 @@ public class CreateEventFragment extends BaseBottomSheetFragment<CreateEventPres
 
     @Override
     public void close() {
-        getActivity().finish();
+        getFragmentManager().popBackStack();
+        ViewUtils.hideKeyboard(binding.getRoot());
     }
 
     @Override
@@ -279,5 +269,11 @@ public class CreateEventFragment extends BaseBottomSheetFragment<CreateEventPres
     private void showLocationLayouts() {
         binding.form.layoutSearchableLocation.setVisibility(View.VISIBLE);
         binding.form.layoutLocationName.setVisibility(View.VISIBLE);
+    }
+
+    @Override
+    public void onStop() {
+        super.onStop();
+        ((AppCompatActivity)getActivity()).getSupportActionBar().show();
     }
 }
--- a/app/src/main/java/org/fossasia/openevent/app/core/main/MainActivity.java
+++ b/app/src/main/java/org/fossasia/openevent/app/core/main/MainActivity.java
@@ -78,6 +78,8 @@ public class MainActivity extends BaseInjectActivity<MainPresenter> implements N
             binding.drawerLayout.closeDrawer(GravityCompat.START);
         } else if (fragmentNavigator.isDashboardActive()) {
             backPressHandler.onBackPressed(this, super::onBackPressed);
+        } else if (getSupportFragmentManager().findFragmentByTag(getResources().getString(R.string.create_event_fragment)) != null) {
+            getSupportFragmentManager().popBackStack();
         } else {
             fragmentNavigator.back();
             binding.navView.getMenu().findItem(R.id.nav_dashboard).setChecked(true);
--- a/app/src/main/java/org/fossasia/openevent/app/common/di/module/ChangeListenerModule.java
+++ b/app/src/main/java/org/fossasia/openevent/app/common/di/module/ChangeListenerModule.java
@@ -4,6 +4,7 @@ import org.fossasia.openevent.app.data.attendee.Attendee;
 import org.fossasia.openevent.app.data.copyright.Copyright;
 import org.fossasia.openevent.app.data.db.DatabaseChangeListener;
 import org.fossasia.openevent.app.data.db.DbFlowDatabaseChangeListener;
+import org.fossasia.openevent.app.data.event.Event;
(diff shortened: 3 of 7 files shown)


iamareebjamal:
@sridharjajoo Status of this PR?

sridharjajoo:
@iamareebjamal Finding difficulty in resolving the issue.

AkshayCHD:
@sridharjajoo what difficulty are you facing?

sridharjajoo:
@iamareebjamal I have solved the issue locally on a different branch. Can I send another P.R for it as the codebase has undergone many changes since this PR was created and there are many complex conflicts?

sridharjajoo:
device-2018-05-09-234420 3
@iamareebjamal Is this fine?

app/src/main/java/org/fossasia/openevent/app/core/event/create/CreateEventFragment.java
```diff
@@ -118,6 +108,7 @@ public void onStart() {
         getPresenter().attach(this);
         binding.setEvent(getPresenter().getEvent());
         getPresenter().start();
+        ((AppCompatActivity)getActivity()).getSupportActionBar().hide();
```
iamareebjamal: Why this?

app/src/main/java/org/fossasia/openevent/app/core/event/create/CreateEventFragment.java
```diff
@@ -280,4 +271,12 @@ private void showLocationLayouts() {
         binding.form.layoutSearchableLocation.setVisibility(View.VISIBLE);
         binding.form.layoutLocationName.setVisibility(View.VISIBLE);
     }
+
+    @Override
+    public void onStop() {
+        super.onStop();
+        ((AppCompatActivity)getActivity()).getSupportActionBar().show();
+        InputMethodManager inputMethodManager = (InputMethodManager) getActivity().getSystemService(INPUT_METHOD_SERVICE);
+        inputMethodManager.hideSoftInputFromWindow(getActivity().getCurrentFocus().getWindowToken(), 0);
```
iamareebjamal: Why this?

app/src/main/java/org/fossasia/openevent/app/core/main/MainActivity.java
```diff
@@ -78,6 +78,8 @@ public void onBackPressed() {
             binding.drawerLayout.closeDrawer(GravityCompat.START);
         } else if (fragmentNavigator.isDashboardActive()) {
             backPressHandler.onBackPressed(this, super::onBackPressed);
+        } else if (getSupportFragmentManager().getBackStackEntryCount() == 2) {
+            getSupportFragmentManager().popBackStack();
```
iamareebjamal: This is magic case. We don't want this. Why is this special case needed at all? All of this should be handled in fragmentNavigator

sridharjajoo: This special case was needed because when we I am adding the CreateEventFragment to backstack and on pressing the back button from this fragment, the dashboard would come up instead of the EventsListFragment because of this code in the onBackPressed( ) 

`  binding.navView.getMenu().findItem(R.id.nav_dashboard).setChecked(true);
            getSupportActionBar().setTitle(R.string.dashboard);
`

app/src/main/java/org/fossasia/openevent/app/core/event/create/CreateEventFragment.java
sridharjajoo: The soft keyboard was persisting after the event had been created. So used this to hide it.

app/src/main/java/org/fossasia/openevent/app/core/event/create/CreateEventFragment.java
sridharjajoo: Actually I wanted the action bar to be hidden so that there could be no navigation possible directly from this CreateEventFragment , hence I hid it.

app/src/main/java/org/fossasia/openevent/app/core/main/MainActivity.java
iamareebjamal: But this is very risky case and hence will cause bugs in future, the fragmentNavigator should already pop backstack until it is empty. The entryCount == 2 has no logical reasoning to go with the event create fragment. Either this needs to be handled by fragmentNavigator transparently, or another solution is needed

app/src/main/java/org/fossasia/openevent/app/core/event/create/CreateEventFragment.java
iamareebjamal: But it should be possible and there should be a back button as well

app/src/main/java/org/fossasia/openevent/app/core/event/create/CreateEventFragment.java
iamareebjamal: I think there is a util method for it and even so doing it in onStop will hide keyboard on rotation

app/src/main/java/org/fossasia/openevent/app/core/main/MainActivity.java
sridharjajoo: What if I add the following to the back( ) of FragmentNaigator

       if (fragmentManager.getBackStackEntryCount() > 1) {
           fragmentManager.popBackStack(); 
           dashboardActive = false;
        } else {
            fragmentManager.popBackStack();
            lastSelectedNavItemId = R.id.nav_dashboard;
            dashboardActive = true;
        }
    }

app/src/main/java/org/fossasia/openevent/app/core/event/create/CreateEventFragment.java
sridharjajoo: Ok. I'll use the method from utility and add it to the CreateEventFragment's close method.

app/src/main/java/org/fossasia/openevent/app/core/main/MainActivity.java
iamareebjamal:
```Java
if (fragmentManager.getBackStackEntryCount() > 1) {
       fragmentManager.popBackStack();
```
This should already include 

`if (fragmentManager.getBackStackEntryCount() == 2)`

iamareebjamal: And that logic will unconditionally pop backstack

sridharjajoo: Ok.I'll look for another solution then.

sridharjajoo: Can I create an Activity for Events List instead of the current EventListFragment and then use the CreateEventFragment in this Activity? Then this fragment will become separate from the Dashboard

iamareebjamal: No, this will become very complex. Also, we are pushing towards having only one activity

sridharjajoo: I have added a TAG like the following when the CreateNewFragment is added

public void openCreateEventFragment() {

                            .add(R.id.fragment_container, CreateEventFragment.newInstance(), "CreateEventFragment")
                            .addToBackStack(null)
                            .commit();
    }

and in the onBackPressed( ) of MainActivity added the following condition:

    public void onBackPressed() {

        if (binding.drawerLayout.isDrawerOpen(GravityCompat.START)) {
            binding.drawerLayout.closeDrawer(GravityCompat.START);
        } else if (fragmentNavigator.isDashboardActive()) {
            backPressHandler.onBackPressed(this, super::onBackPressed);
        } **else if (getSupportFragmentManager().findFragmentByTag("CreateEventFragment") != null) {
            getSupportFragmentManager().popBackStack();
        }** else {
            fragmentNavigator.back();
            binding.navView.getMenu().findItem(R.id.nav_dashboard).setChecked(true);
            getSupportActionBar().setTitle(R.string.dashboard);
        }
    }

I guess this will be a safer operation.

iamareebjamal: This is unneeded dynamic check. This is not a good solution because it is not general. This is relying too much on checks like this. This will surely break if behaviour of navigator is changed. You'll have to add 5 more such checks when new fragments for session, speaker and other details are added, So saying this, it is a sloppy workaround, not a fix. I know you might be thinking if it is working why not just use it, because this will solve one issue and create 100 more in future.

The error you are facing may be solved by adding 2-3 lines of 'correct' logical code in fragment navigator. This is because one of the things is broken and we are patching wrong stuff. I don't know what is broken, or would have told you so, but this is definitely not the right solution

Any kind of dynamic check on CreateEventFragment is a no go. This should be a general fix

PR #874, opened 2018-04-05
feat: Show Tracks under an Event

### LINKED ISSUE #872: Show Tracks under an Event
### Feature Request

**Feature Description**

Add a TracksFragment listing all the Tracks associated with the Event.

Yes

**Parent issue**

#870

srv-twry:
Fixes #872 

Changes: Implmented Tracks module
Fetched Tracks for an Event
Saved it in the database for offline use
Show it in a RecyclerView

Screenshots for the change: 
videotogif_2018 04 05_22 53 57

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/app/core/track/list/TracksFragment.java
+++ b/app/src/main/java/org/fossasia/openevent/app/core/track/list/TracksFragment.java
@@ -1,0 +1,143 @@
+package org.fossasia.openevent.app.core.track.list;
+
+import android.content.Context;
+import android.databinding.DataBindingUtil;
+import android.os.Bundle;
+import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
+import android.support.v4.widget.SwipeRefreshLayout;
+import android.support.v7.widget.DefaultItemAnimator;
+import android.support.v7.widget.LinearLayoutManager;
+import android.support.v7.widget.RecyclerView;
+import android.view.LayoutInflater;
+import android.view.View;
+import android.view.ViewGroup;
+
+import org.fossasia.openevent.app.R;
+import org.fossasia.openevent.app.common.mvp.view.BaseFragment;
+import org.fossasia.openevent.app.core.main.MainActivity;
+import org.fossasia.openevent.app.data.tracks.Track;
+import org.fossasia.openevent.app.databinding.TracksFragmentBinding;
+import org.fossasia.openevent.app.ui.ViewUtils;
+
+import java.util.List;
+
+import javax.inject.Inject;
+
+import dagger.Lazy;
+
+public class TracksFragment extends BaseFragment<TracksPresenter> implements TracksView {
+    private Context context;
+    private long eventId;
+
+    @Inject
+    Lazy<TracksPresenter> tracksPresenter;
+
+    private TracksAdapter tracksAdapter;
+    private TracksFragmentBinding binding;
+    private SwipeRefreshLayout refreshLayout;
+
+    private boolean initialized;
+
+    public static TracksFragment newInstance(long eventId) {
+        TracksFragment fragment = new TracksFragment();
+        Bundle args = new Bundle();
+        args.putLong(MainActivity.EVENT_KEY, eventId);
+        fragment.setArguments(args);
+        return fragment;
+    }
+
+    @Override
+    public void onCreate(@Nullable Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+
+        context = getContext();
+        if (getArguments() != null)
+            eventId = getArguments().getLong(MainActivity.EVENT_KEY);
+    }
+
+    @Nullable
+    @Override
+    public View onCreateView(@NonNull LayoutInflater inflater, @Nullable ViewGroup container, @Nullable Bundle savedInstanceState) {
+        binding = DataBindingUtil.inflate(inflater, R.layout.tracks_fragment, container, false);
+
+        binding.createTrackFab.setOnClickListener(view -> {
+            // Create Track
+        });
+
+        return binding.getRoot();
+    }
+
+    @Override
+    public void onStart() {
+        super.onStart();
+        setupRecyclerView();
+        setupRefreshListener();
+        getPresenter().attach(eventId, this);
+        getPresenter().start();
+
+        initialized = true;
+    }
+
+    @Override
+    protected int getTitle() {
+        return R.string.tracks;
+    }
+
+    @Override
+    public void onStop() {
+        super.onStop();
+        refreshLayout.setOnRefreshListener(null);
+    }
+
+    private void setupRecyclerView() {
+        if (!initialized) {
+            tracksAdapter = new TracksAdapter(getPresenter());
+
+            RecyclerView recyclerView = binding.tracksRecyclerView;
+            recyclerView.setLayoutManager(new LinearLayoutManager(context));
+            recyclerView.setAdapter(tracksAdapter);
+            recyclerView.setItemAnimator(new DefaultItemAnimator());
+        }
+    }
+
+    private void setupRefreshListener() {
+        refreshLayout = binding.swipeContainer;
+        refreshLayout.setColorSchemeColors(getResources().getColor(R.color.color_accent));
+        refreshLayout.setOnRefreshListener(() -> {
+            refreshLayout.setRefreshing(false);
+            getPresenter().loadTracks(true);
+        });
+    }
+
+    @Override
+    public void showError(String error) {
+        ViewUtils.showSnackbar(binding.getRoot(), error);
+    }
+
+    @Override
+    public void showProgress(boolean show) {
+        ViewUtils.showView(binding.progressBar, show);
+    }
+
+    @Override
+    public void onRefreshComplete(boolean success) {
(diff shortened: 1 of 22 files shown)


Masquerade0097:
I think it would be better if you add the Track Table for Database Migration as well.

srv-twry:
@Masquerade0097  How to add it for migration ? Would adding it to the recreated Array 
```Class[] recreated = new Class[] {Faq.class, Copyright.class};``` work or would i have to bump up the database version to 10 ?

**Comment by Masquerade0097** [comment]
@srv-twry Creating version 10 would be fine.

**Comment by iamareebjamal** [comment]
Very nice, but use the color and font color of the track in card header to make it look nice

**Comment by srv-twry** [comment]
screenshot_2018-04-06-22-53-07-75

File: app/src/main/res/layout/tracks_item.xml [hunk 1]
```diff
@@ -39,7 +40,7 @@
                     android:layout_marginLeft="@dimen/spacing_small"
                     android:layout_marginStart="@dimen/spacing_small"
                     android:text='@{ track.name }'
-                    android:textColor="@android:color/black"
+                    android:textColor='@{ track.fontColor != null ? Color.parseColor(track.fontColor) : Color.BLACK }'
```
iamareebjamal: Color.parseColor can crash if the color is not in the correct format. Get the right formats from presenter with defensive checks

app/src/main/java/org/fossasia/openevent/app/core/track/list/viewholder/TracksViewHolder.java
```diff
@@ -0,0 +1,31 @@
...
+        if (!validColor(track.fontColor)) {
+            track.fontColor = null;
+        }
+        binding.setTrack(track);
+        binding.executePendingBindings();
+    }
+
+    private boolean validColor(String color) {
+        String pattern = "^#[0-9A-Fa-f]+$";
+        return color.matches(pattern) && (color.length() == 7 || color.length() == 9);
```
srv-twry: @iamareebjamal The official documentation [here](https://developer.android.com/reference/android/graphics/Color.html#parseColor%28java.lang.String%29) states that ```Color.parseColor``` expects colors in ```#RRGGBB```
```#AARRGGBB``` format.
Hence i am using regular expressions to check if the color is one of those. If it isn't then setting it to null, so that default value will be used. Is it okay ?

Comment (iamareebjamal) [reply app/src/main/java/org/fossasia/openevent/app/core/track/list/viewholder/TracksViewHolder.java]: Rather than checking, do exception handling. Set header to light grey and text to black if there's any error

Comment (iamareebjamal) [reply app/src/main/java/org/fossasia/openevent/app/core/track/list/viewholder/TracksViewHolder.java]: Rather than coloring whole card, only color the header

Comment (srv-twry) [reply app/src/main/java/org/fossasia/openevent/app/core/track/list/viewholder/TracksViewHolder.java]: I am facing an issue here
This is my bind method

``` Java
public void bind(Track track) {
        try {
            track.color = "#nmnmnnm";
            binding.setTrack(track);
            binding.executePendingBindings();
        } catch (IllegalArgumentException illegalArgumentException) {
            track.color = null;
            track.fontColor = null;
        }
    }
```

I am trying to test the error and hence setting the ```track.color``` variable to an illegal argument. The exception is catched and then i set the color and fontColor to null which would mean that the layout file will use the default values which i have mentioned in it.
The problem is i am getting an empty list. 

Screenshot: 
 screenshot_2018-04-07-19-08-34-65

I can't figure out the source of this problem.

iamareebjamal: Because execute pending binding is not called in case of an exception

srv-twry: Problem persists even after changing the ```bind``` method to the following
``` Java
public void bind(Track track) {
        try {
            track.color = "#nmnmnnm";
            binding.setTrack(track);
            binding.executePendingBindings();
        } catch (IllegalArgumentException illegalArgumentException) {
            track.color = null;
            track.fontColor = null;
            binding.setTrack(track);
            binding.executePendingBindings();
        }
    }
```

or 

``` Java
public void bind(Track track) {
        try {
            track.color = "#nmnmnnm";
            binding.setTrack(track);
            binding.executePendingBindings();
        } catch (IllegalArgumentException illegalArgumentException) {
            track.color = null;
            track.fontColor = null;
            binding.executePendingBindings();
        }
    }
```

iamareebjamal: BTW, you shouldn't do this at all. Create two methods in presenter which take a track as an argument and return a color. This problem will be solved

srv-twry:
@iamareebjamal Ready for review.
screenshot_2018-04-08-15-13-07-94

iamareebjamal:
The header should be match parent. Take a look at the sessions layout in Open Event Android

srv-twry:
screenshot_2018-04-08-15-56-04-60

srv-twry:
@iamareebjamal Please review.

app/src/main/res/layout/tracks_item.xml
```diff
@@ -0,0 +1,63 @@
...
+
+                <TextView
+                    android:layout_width="match_parent"
+                    android:layout_height="wrap_content"
+                    android:layout_marginLeft="@dimen/spacing_small"
+                    android:layout_marginStart="@dimen/spacing_small"
+                    android:text='@{ track.name }'
+                    android:textColor='@{ viewHolder.getTrackFontColor(track) }'
+                    android:textStyle="bold"
+                    android:textSize="@dimen/text_size_large" />
```
iamareebjamal: Instead use `style=AppCompat.Text.Header`

app/src/main/res/layout/tracks_item.xml
```diff
@@ -0,0 +1,63 @@
+<?xml version="1.0" encoding="utf-8"?>
+<layout xmlns:android="http://schemas.android.com/apk/res/android">
+
+    <data>
+
+        <import type="android.view.View" />
+
+        <variable
+            name="viewHolder"
+            type="org.fossasia.openevent.app.core.track.list.viewholder.TracksViewHolder" />
```
iamareebjamal: This won't be needed when you will have static method. And instead of adding those methods in ViewHolder, add these methods in Presenter

app/src/main/res/layout/tracks_item.xml
```diff
@@ -0,0 +1,63 @@
...
+
+            </LinearLayout>
+
+            <TextView
+                android:layout_width="match_parent"
+                android:layout_height="wrap_content"
+                android:layout_marginLeft="@dimen/spacing_small"
+                android:layout_marginStart="@dimen/spacing_small"
+                android:text='@{ track.description }'
+                android:textSize="@dimen/text_size_normal"
```
iamareebjamal: style Content from AppCompat or something

app/src/main/res/layout/tracks_item.xml
```diff
@@ -0,0 +1,63 @@
...
+            android:foreground="?selectableItemBackground"
+            android:gravity="center_vertical"
+            android:orientation="vertical">
+
+            <LinearLayout
+                android:layout_width="match_parent"
+                android:layout_height="wrap_content"
+                android:background="@{ viewHolder.getTrackColor(track) }"
+                android:orientation="vertical"
+                style="@style/ItemPadding">
```
iamareebjamal: This LinearLayout is useless. Instead just have the below header textview take up all the space for this and add background color to that

app/src/main/java/org/fossasia/openevent/app/core/track/list/viewholder/TracksViewHolder.java
```diff
@@ -0,0 +1,42 @@
...
+
+    public int getTrackFontColor(Track track) {
+        int color;
+        try {
+            color = Color.parseColor(track.fontColor);
+        } catch (Exception exception) {
+            color = Color.BLACK;
+        }
+        return color;
+    }
```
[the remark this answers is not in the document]
srv-twry: Please explain the meaning of adding tests ? Did you mean adding unit tests for this ? I intend to add Unit tests for all these features in a separate PR.

app/src/main/java/org/fossasia/openevent/app/core/track/list/TracksFragment.java
```diff
@@ -0,0 +1,146 @@
...
+import javax.inject.Inject;
+
+import dagger.Lazy;
+
+public class TracksFragment extends BaseFragment<TracksPresenter> implements TracksView {
+    private Context context;
+    private long eventId;
+
+    @Inject
+    ContextUtils utilModel;
```
iamareebjamal: Not to be used in Fragments

srv-twry:
screenshot_2018-04-08-23-57-31-02

iamareebjamal:
Description should have padding and should not appear to be bold

srv-twry:
screenshot_2018-04-09-00-23-58-84

srv-twry:
@iamareebjamal Please review

PR #875, opened 2018-04-06
feat: Select and Delete FAQ from List

### LINKED ISSUE #728: Add option to delete FAQ
**Is it a ?**

- [ ] Bug Report

- [x] Feature Request
    
- [ ] Chore
    
**Parent Issue**
#652 

**Description**

There should be an option in FAQ List Layout to delete an FAQ.

Yes

Masquerade0097:
Fixes #728 

Changes:   Following feature set is added.
1. User can select(by long click) any FAQ from list to delete it from toolbar.
2. If other list item is long clicked then it get selected instead of the first one.
3. If user clicks other list item while an item is already selected then it gets deselected as well as if the user after selecting an item, clicks it again then it gets deselected.
4. If a list item is selected and user clicks to create new FAQ then it gets deselected.

Screenshots for the change:   Below are the screen records for the changes mentioned above.
1. 
del1

2 & 3.
del2

4.
del3


app/src/main/java/org/fossasia/openevent/app/core/faq/list/FaqListFragment.java
```diff
@@ -117,6 +129,57 @@ private void setupRefreshListener() {
...
+                if (deleteDialog == null)
+                    deleteDialog = new AlertDialog.Builder(context)
+                        .setTitle(R.string.delete_confirmation)
+                        .setMessage(R.string.delete_confirmation_message)
+                        .setPositiveButton(R.string.ok, (dialog, which) -> getPresenter().deleteSelectedFaq())
+                        .setNegativeButton(R.string.cancel, (dialog, which) -> dialog.dismiss())
+                        .create();
+
+                deleteDialog.show();
+                deletingMode = false;
```
iamareebjamal: Move to its own method

app/src/main/java/org/fossasia/openevent/app/core/faq/list/FaqListFragment.java
```diff
@@ -117,6 +129,57 @@ private void setupRefreshListener() {
...
+    }
+
+    @Override
+    public void onPrepareOptionsMenu(Menu menu) {
+        super.onPrepareOptionsMenu(menu);
+        if (deletingMode) {
+            MenuItem menuItem = menu.findItem(R.id.del);
+            menuItem.setVisible(true);
+        } else {
+            MenuItem menuItem = menu.findItem(R.id.del);
```
iamareebjamal: This line is common, can be moved out of conditional

app/src/main/java/org/fossasia/openevent/app/core/faq/list/FaqListFragment.java
```diff
@@ -117,6 +129,57 @@ private void setupRefreshListener() {
...
+            menuItem.setVisible(false);
```
iamareebjamal: Can be reduced to 
```Java
menuItem.setVisible(deletingMode);
```

app/src/main/res/values/strings.xml
```diff
@@ -194,6 +194,8 @@
     <string name="welcome">Hi, Organiser!</string>
     <string name="delete_copyright">Delete Copyright</string>
     <string name="year_validation_error">Please Enter a Valid Year</string>
+    <string name="delete_confirmation_message">Are you sure you want to delete the Question?</string>
+    <string name="delete_confirmation">Delete</string>
```
iamareebjamal: Can be reduced to `delete`

app/src/main/java/org/fossasia/openevent/app/core/faq/list/FaqListFragment.java
```diff
@@ -117,6 +129,59 @@ private void setupRefreshListener() {
...
+    }
+
+    @Override
+    public void onPrepareOptionsMenu(Menu menu) {
+        super.onPrepareOptionsMenu(menu);
+        MenuItem menuItem = menu.findItem(R.id.del);
+        if (deletingMode)
+            menuItem.setVisible(deletingMode);
+        else
+            menuItem.setVisible(deletingMode);
```
iamareebjamal: no need of if else here

app/src/main/java/org/fossasia/openevent/app/core/faq/list/FaqListPresenter.java
```diff
@@ -75,4 +81,43 @@ private void listenChanges() {
     public List<Faq> getFaqs() {
         return faqs;
     }
+
+    public void deleteFaqWithId(long id, boolean forceReload) {
```
iamareebjamal: Is there any need of forceReload parameter here

app/src/main/java/org/fossasia/openevent/app/core/faq/list/FaqListPresenter.java
```diff
@@ -75,4 +81,43 @@ private void listenChanges() {
...
+                getView().showMessage("FAQ Deleted. Refreshing Items");
+                loadFaqs(forceReload);
+            }, Logger::logError);
+    }
+
+    public void deleteSelectedFaq() {
+        deleteFaqWithId(faqId, true);
+    }
+
+    public void selectSpecificFaq(List<Faq> faqs, Long faqId, boolean selection) {
```
iamareebjamal: What does this do?

app/src/main/java/org/fossasia/openevent/app/core/faq/list/viewholder/FaqViewHolder.java
```diff
@@ -5,20 +5,45 @@
...
+import org.fossasia.openevent.app.common.Pipe;
 import org.fossasia.openevent.app.data.models.Faq;
 import org.fossasia.openevent.app.databinding.FaqLayoutBinding;
 
 public class FaqViewHolder extends RecyclerView.ViewHolder {
 
     private final FaqLayoutBinding binding;
+    private Faq faq;
+    private Pipe<Long> longClickAction;
+    private Pipe<Long> clickAction;
```
iamareebjamal: Use `Function`

app/src/main/java/org/fossasia/openevent/app/core/faq/list/FaqListPresenter.java
Masquerade0097: When a list item is selected and then the layout is swipe refreshed then I need to deselect all the list items. Also when an item is selected and then other item is long pressed then I need to deselect all the list items other than that. The faqId method parameter contains the id of the list item which needs to be selected. This method address to the above two cases under if and else respectively.

app/src/main/java/org/fossasia/openevent/app/core/faq/list/FaqListPresenter.java
Masquerade0097: Ya, it's not necessary. I'll remove it.

app/src/main/java/org/fossasia/openevent/app/core/faq/list/viewholder/FaqViewHolder.java
Masquerade0097: The return type of referenced methods like `toolbarDeleteMode()` and `listItemClickAction` is void. So why is `Function` needed instead of `Pipe` ?

app/src/main/java/org/fossasia/openevent/app/core/faq/list/FaqListPresenter.java
Masquerade0097: On rotation, the ticket Fragment, FAQ Fragment, Attendee Fragment, Event Fragment all gets destroyed and user lands on Dashboard. So I think persisting the fragment on screen rotation should be ensured first as a separate issue. What do you say?

app/src/main/java/org/fossasia/openevent/app/core/faq/list/viewholder/FaqViewHolder.java
iamareebjamal: Yes, it shouldn't be Function, it should be runnable

app/src/main/java/org/fossasia/openevent/app/core/faq/list/FaqListPresenter.java
iamareebjamal: Yes, alright

Masquerade0097:
@iamareebjamal Now, I'm storing 2 faq objects in the presenter. First,  `currentFaq` - the Faq which is selected in the list and second,`previousFaq` - the Faq which was selected before the current Faq. Then when i need to deselect all the Faq then rather than traversing the entire list I will just unselect these 2 Faqs. Also this helps in switching (selecting) from one Faq to another by remembering previous Faq. 

Further, this don't need to traverse along the list so maybe we won't need to use runnable to do that process in a different thread. Is it okay? I will remove merge conflicts and update the unit tests acordingly.

app/src/main/java/org/fossasia/openevent/app/core/faq/list/FaqListPresenter.java
```diff
@@ -18,14 +18,18 @@
...
+import static org.fossasia.openevent.app.common.rx.ViewTransformers.progressiveErroneousCompletable;
 import static org.fossasia.openevent.app.common.rx.ViewTransformers.progressiveErroneousRefresh;
 
 public class FaqListPresenter extends BaseDetailPresenter<Long, IFaqListView> {
 
     private final List<Faq> faqs = new ArrayList<>();
     private final IFaqRepository faqRepository;
     private final IDatabaseChangeListener<Faq> faqChangeListener;
+    private Faq previousFaq = new Faq();
+    private Faq currentFaq = new Faq();
```
iamareebjamal: Why allocate new objects?

iamareebjamal: Why do we need a previous faq object. Just compare the incoming one with the current stored one

app/src/main/java/org/fossasia/openevent/app/core/faq/list/viewholder/FaqViewHolder.java
```diff
@@ -5,20 +5,45 @@
...
+import org.fossasia.openevent.app.common.Pipe;
 import org.fossasia.openevent.app.data.models.Faq;
 import org.fossasia.openevent.app.databinding.FaqLayoutBinding;
 
 public class FaqViewHolder extends RecyclerView.ViewHolder {
 
     private final FaqLayoutBinding binding;
+    private Faq faq;
+    private Pipe<Faq> longClickAction;
+    private Pipe<Long> clickAction;
```
iamareebjamal: Make this Runnable

PR #889, opened 2018-04-11
Fix bug: Pressing back on create/edit copyright exits activity #884

### LINKED ISSUE #884: Pressing back on create/edit copyright exits activity
### Bug Report    

**Actual Behaviour**

When pressing back either via soft or hard button, the about event activity is finished.

**Expected Behaviour**

The fragment should be popped from backstack instead.

No

mishuvs:
Fixes #884 

Changes: 
    1. Add Copyright Fragment to backstack
    2. Change activity's home action to `finish()` only when there's no fragment in backstack

Screenshots of the change: 
backstackcopyright

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/app/core/event/about/AboutEventActivity.java
+++ b/app/src/main/java/org/fossasia/openevent/app/core/event/about/AboutEventActivity.java
@@ -2,6 +2,7 @@ package org.fossasia.openevent.app.core.event.about;
 
 import android.os.Bundle;
 import android.support.v4.app.Fragment;
+import android.support.v4.app.FragmentManager;
 import android.support.v7.app.AppCompatActivity;
 import android.view.MenuItem;
 
@@ -17,6 +18,8 @@ public class AboutEventActivity extends AppCompatActivity implements HasSupportF
 
     public static final String EVENT_ID = "event_id";
 
+    private final FragmentManager fragmentManager = getSupportFragmentManager();
+
     @Inject
     DispatchingAndroidInjector<Fragment> dispatchingAndroidInjector;
 
@@ -27,8 +30,8 @@ public class AboutEventActivity extends AppCompatActivity implements HasSupportF
 
         Bundle extras = getIntent().getExtras();
         if (extras != null && extras.containsKey(EVENT_ID) && savedInstanceState == null) {
-            getSupportFragmentManager().beginTransaction()
-                .replace(R.id.fragment, AboutEventFragment.newInstance(extras.getLong(EVENT_ID)))
+            fragmentManager.beginTransaction()
+                .add(R.id.fragment, AboutEventFragment.newInstance(extras.getLong(EVENT_ID)))
                 .commit();
         }
     }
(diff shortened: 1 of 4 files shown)


mishuvs:
@iamareebjamal Please review.

app/src/main/java/org/fossasia/openevent/app/core/event/about/AboutEventActivity.java
```diff
@@ -37,7 +37,10 @@ protected void onCreate(Bundle savedInstanceState) {
     public boolean onOptionsItemSelected(MenuItem item) {
         switch (item.getItemId()) {
             case android.R.id.home:
-                finish();
+                if (getSupportFragmentManager().getBackStackEntryCount() > 0)
+                    getSupportFragmentManager().popBackStack();
+                else
+                    finish();
```
iamareebjamal: Clicking on hardware back button should work as well

app/src/main/java/org/fossasia/openevent/app/core/event/about/AboutEventFragment.java
```diff
@@ -125,11 +125,13 @@ public boolean onOptionsItemSelected(MenuItem item) {
             case R.id.action_create_change_copyright:
                 if (creatingCopyright) {
                     getFragmentManager().beginTransaction()
-                        .replace(R.id.fragment, CreateCopyrightFragment.newInstance())
+                        .add(R.id.fragment, CreateCopyrightFragment.newInstance())
```
iamareebjamal: Don't use add

app/src/main/java/org/fossasia/openevent/app/core/event/about/AboutEventActivity.java
mishuvs: It does. Since the fragment is added to backstack, clicking phone's back button works as well by reversing the transition.

iamareebjamal: OK, use a shared variable for FragmentManager

app/src/main/java/org/fossasia/openevent/app/core/event/about/AboutEventFragment.java
mishuvs: Using `#add` will retain AboutEventFragment's state. So on back-press, no progress bar will be shown, making it look smooth. The newly created copyright is added instantly to event details without a progress bar. What do you think?

app/src/main/java/org/fossasia/openevent/app/core/event/about/AboutEventActivity.java
mishuvs: Umm, it's used only once. Why use a shared variable?

app/src/main/java/org/fossasia/openevent/app/core/event/about/AboutEventFragment.java
iamareebjamal: It takes more memory but alright, we'll switch to add if there are more fragments

mishuvs: Is that a yes or do we add `add` later?

PR #893, opened 2018-04-12
fix: Pressing back pops from backstack

### LINKED ISSUE #885: Pressing back on edit organizer details exits activity
### Bug Report    

**Actual Behaviour**

When pressing back either via soft or hard button, the about event organizer is finished.

**Expected Behaviour**

The fragment should be popped from backstack instead.

No

sridharjajoo:
Fixes #885 

Changes:
The edit organizer info fragment was added to backstack and the onBackPressed( ) was overridden so that when the count of number of fragment in backstack is greater than one , then the fragment should pop from the backstack and if count is 0 , activity can be exited.

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/app/core/organizer/detail/OrganizerDetailActivity.java
+++ b/app/src/main/java/org/fossasia/openevent/app/core/organizer/detail/OrganizerDetailActivity.java
@@ -2,6 +2,7 @@ package org.fossasia.openevent.app.core.organizer.detail;
 
 import android.os.Bundle;
 import android.support.v4.app.Fragment;
+import android.support.v4.app.FragmentManager;
 import android.support.v7.app.AppCompatActivity;
 import android.view.MenuItem;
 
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/fossasia/openevent/app/core/organizer/detail/OrganizerDetailActivity.java
```diff
@@ -45,4 +45,19 @@ public boolean onOptionsItemSelected(MenuItem item) {
     public AndroidInjector<Fragment> supportFragmentInjector() {
         return dispatchingAndroidInjector;
     }
+
+    @Override
+    public void onBackPressed() {
```
iamareebjamal: Look at the PR merged by @mishuvs

app/src/main/java/org/fossasia/openevent/app/core/organizer/detail/OrganizerDetailActivity.java
```diff
@@ -45,4 +45,19 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
+        handleBackPress();
+    }
+
+    private void handleBackPress() {
+
```
iamareebjamal: Store fragment manager in a global variable

PR #907, opened 2018-04-28
fix: Toolbar goes behind the status bar on login screen

### LINKED ISSUE #681: Toolbar goes behind the status bar on login screen

brijeshshah13:
Fixes #681.

Changes: Added `android:fitsSystemWindows="true"` to the root layout of the login screen to resolve the issue.

Screenshots for the change: 

screenshot_20180428-121148

@iamareebjamal Please review.


iamareebjamal:
Where's the toolbar gone?

brijeshshah13:
@iamareebjamal It will minimize only when the user is entering the `Base URL`.

iamareebjamal:
It shouldn't get minimized

brijeshshah13:
@iamareebjamal I think as we are not using `AppBarLayout` in the `login_fragment.xml` and setting the title of the fragment programmatically, we won't be able to stick the title to the top. Also, when the `NestedScrollView` is scrolled, it still affects the title.

When I scroll to the bottom and then click on the Base URL, still the title overlaps with the status bar:

screenshot_20180428-170033

I think using `AppBarLayout` will solve this problem. However, to add that we need to remove the current implementation of toolbar. How exactly are we adding it again, because i can't seem to locate from where the `Log In` toolbar is defined in the xml.

iamareebjamal:
Look in auth activity xml

brijeshshah13:
I'm asking about the `eventyay organizer` toolbar in the following screenshot:

capture

brijeshshah13:
@iamareebjamal Please attend this PR at your comfort.

iamareebjamal:
The text i changed through Java, I don't understand what problem you are facing?

brijeshshah13:
@iamareebjamal I just wanted to make sure that the toolbar for the `login_fragment.xml` is indeed actually of the `main_app_bar.xml`.

iamareebjamal:
Yes that would be so, but to confirm, you could just look at the actual login fragment, implementation and root layout

brijeshshah13:
@iamareebjamal That is where m confused. I can't figure how is it imported from the `main_app_bar.xml`. Also, i would be glad if you could be help me with some pointers. I tried many solutions with negative results.

brijeshshah13:
@iamareebjamal I'm pretty sure the issue is resolved now. Here is a GIF to demonstrate the same:

app/src/main/AndroidManifest.xml
```diff
@@ -22,7 +22,8 @@
         <activity
             android:name=".core.auth.AuthActivity"
             android:theme="@style/AppTheme.Main.Light"
-            android:label="${appName}">
+            android:label="${appName}"
+            android:windowSoftInputMode="stateHidden|adjustResize">
```
iamareebjamal: This is restricting the layout to not re-render when adjusted or resized which will create problems when using tab layout or dual pane mode

brijeshshah13: Can you please specify some conditions where it is raising the issues you mentioned which I can test?

iamareebjamal: Currently, there is no problem. In future, there may be. We can't just enable these flags and change the resize behaviour just to fix a bug which is not caused by them. This is a workaround, not a fix

brijeshshah13: I tested it with the dual pane mode, I think it is working fine.

brijeshshah13: I just now tested it with dual pane mode, all looks good to me. And this behavior will only affect the layouts which are inherited by the AuthActivity. You can verify that the current approach is not just a workaround from the [documentation](https://developer.android.com/training/keyboard-input/visibility). Still, if you point me the exact issues, I can try to address them as well.

brijeshshah13: Adding to the description of what's been done, `adjustResize` will only ensure that the system resizes your layout to the available space and `stateHidden` will only hide the soft keyboard till the user actually clicks on any of the field to input values.

brijeshshah13: Let me know what conclusion you reached.

iamareebjamal: Still, this is a workaround which is not needed for most activities, and not a fix

Add a todo in manifest acknowledging this

brijeshshah13: @iamareebjamal Done.

PR #951, opened 2018-05-16
bug:Added validation to edit text fields

### LINKED ISSUE #950: Bug while entering URL in Create Sponsors Fragment
### Bug Report    

**Actual Behaviour**
While a sponsor is being created and the URL Is being entered then there should be a check simultaneously that whether the URL is being entered is correct or not in correct format or not.

 **Expected Behaviour**
There should be validation while a URL is being entered.
 
@iamareebjamal I would like to work on it

sridharjajoo:
Fixes #950 

Screenshots for the change: 
whatsapp image 2018-05-17 at 00 01 44 1

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/app/core/sponsor/create/CreateSponsorFragment.java
+++ b/app/src/main/java/org/fossasia/openevent/app/core/sponsor/create/CreateSponsorFragment.java
@@ -3,14 +3,20 @@ package org.fossasia.openevent.app.core.sponsor.create;
 import android.databinding.DataBindingUtil;
 import android.os.Bundle;
 import android.support.annotation.Nullable;
+import android.support.design.widget.TextInputLayout;
+import android.text.Editable;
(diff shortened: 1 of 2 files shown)


sridharjajoo:
@iamareebjamal Kindly review

app/src/main/java/org/fossasia/openevent/app/core/sponsor/create/CreateSponsorView.java
```diff
@@ -8,4 +11,5 @@
 
     void dismiss();
 
+    void validate(TextInputLayout textInputLayout, Function<String, Boolean> function, String str);
```
iamareebjamal: Why is there a need to add this?

sridharjajoo: Sorry ,not needed. I'll remove it from interface and add a simple method

PR #954, opened 2018-05-17
chore: Add tests for sponsor module #953

### LINKED ISSUE #953: Write tests for each feature in Sponsor module
**Is it a ?**

- [ ] Bug Report

- [ ] Feature Request
    
- [x] Chore
    
### Chore

**Description**

Create Unit tests for Create sponsor Module.

**Parent issue**
#730 

Yes

mishuvs:
Fixes #953 

Changes: Added tests for SponsorsPresenter and SponsorsRepository


app/src/test/java/org/fossasia/openevent/app/core/presenter/SponsorsPresenterTest.java
```diff
@@ -0,0 +1,179 @@
...
+        when(sponsorRepository.getSponsors(anyLong(), anyBoolean())).thenReturn(Observable.fromIterable(SPONSORS));
+        when(databaseChangeListener.getNotifier()).thenReturn(PublishSubject.create());
+
+        sponsorsPresenter.start();
+
+        verify(sponsorRepository).getSponsors(ID, false);
+    }
+
+    @Test
+    public void shouldShowTicketsAutomatically() {
```
srv-twry: You mean ```shouldShowSponsorsAutomatically```, right ?

mishuvs: I was gonna change, that's why I marked as WIP 😅

mishuvs:
@iamareebjamal Please review

app/src/main/java/org/fossasia/openevent/app/core/sponsor/list/SponsorsPresenter.java
```diff
@@ -97,7 +97,7 @@ public void loadSponsors(boolean forceReload) {
         return sponsors;
     }
 
-    private void deleteSponsor(Long sponsorId) {
+    public void deleteSponsor(Long sponsorId) {
```
iamareebjamal: Should be package protected and with annotation `@VisibleForTesting`

app/src/main/java/org/fossasia/openevent/app/core/sponsor/list/SponsorsPresenter.java
```diff
@@ -172,4 +172,8 @@ private int countSelected() {
         return count;
     }
 
+    public Map<Long, ObservableBoolean> getSelectedSponsors() {
```
iamareebjamal: Same here

app/src/main/java/org/fossasia/openevent/app/core/sponsor/list/SponsorsPresenter.java
mishuvs: It gives `error: getSelectedSponsors() is not public in SponsorsPresenter; cannot be accessed from outside package` when no modifier is used, and `.. is protected...` when protected is used.

iamareebjamal: If you are testing its functionality in some other way, then don't explicitly test it

mishuvs: `deleteSponsor()` is indirectly tested when we test `deleteSelectedSponsors()`. So I guess I should remove its test? But for `getSelectedSponsors()`, the above error persists because of different package. Should I make it public again?

iamareebjamal: You should add test in the same package as presenter

mishuvs:
@iamareebjamal Please review

PR #1009, opened 2018-06-02
Add BottomNavigationView in Events list

### LINKED ISSUE #972: Convert event list to tab layout
**Is it a ?**

- [ ] Bug Report

- [x] Feature Request
    
- [ ] Chore
    
### Feature Request

**Feature Description**

- [ ] Similar to the Eventbrite app, we need to have a tab layout for the event list
- [ ] The event list item should also show general info like date, the percentage of tickets sold, number of attendees Checked In, and revenue generated (#994)

device-2018-05-24-045801

Yes

mishuvs:
Fixes #972

Changes: Adds BottomNavigationView in Events List

Gif for the change:
videotogif_2018 06 03_07 01 28

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/app/core/event/list/EventsListAdapter.java
+++ b/app/src/main/java/org/fossasia/openevent/app/core/event/list/EventsListAdapter.java
@@ -3,6 +3,8 @@ package org.fossasia.openevent.app.core.event.list;
 import android.support.v7.widget.RecyclerView;
 import android.view.LayoutInflater;
 import android.view.ViewGroup;
+import android.widget.Filter;
+import android.widget.Filterable;
 
 import com.timehop.stickyheadersrecyclerview.StickyRecyclerHeadersAdapter;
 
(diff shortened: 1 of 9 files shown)


iamareebjamal:
Great work @mishuvs 

Will be reviewing soon

app/src/main/java/org/fossasia/openevent/app/core/event/list/EventsListAdapter.java
```diff
@@ -13,24 +13,56 @@
...
+                    pastEvents.add(event);
+                else if ("live".equalsIgnoreCase(category))
+                    liveEvents.add(event);
+                else
+                    upcomingEvents.add(event);
+            } catch (ParseException e) {
+                Timber.e(e);
+            }
+        }
+        notifyDataSetChanged();
```
iamareebjamal: Implement this through filter instead

iamareebjamal:
@mishuvs It has been 2 days

app/src/main/java/org/fossasia/openevent/app/core/event/list/EventsListAdapter.java
mishuvs: Sorry, I didn't get this one. AFAIK filter is used to filter items emitted by Observable, but here we want every item to be updated. So why use filter? 
Could you please elaborate a little?
Also I was thinking of moving the methods added to the adapter, to the presenter. In that case we can just update the `events` List every time. But I am stuck on how to listen for its changes.

iamareebjamal: I meant use the adapter filtering feature instead of maintaining 3 lists

iamareebjamal:
Great work @mishuvs

PR #1044, opened 2018-06-09
feat: Extract title and detail of error response

### LINKED ISSUE #1038: Incorrect error message in creating Session/Ticket
**Is it a ?**

- [x] Bug Report

- [ ] Feature Request
    
- [ ] Chore
    
### Bug Report    

**Actual Behaviour**

When sessions, tickets are created for an event for which they are disabled. The error message produced is `You are not authorized to make this request`.

**Expected Behaviour**

The produced error message should be `Sessions are disabled for this event`,`Ticketing is disabled for this event`.

**Steps to reproduce it**

Disable the `is-speakers-sponsors-enabled`, `is-ticketing-enabled` field for an event and then try to create the same.

**LogCat for the issue**
`D/OkHttp: {"errors": [{"status": 403, "source": {"pointer": ""}, "title": "Access Forbidden", "detail": "Sessions are disabled for this Event"}], "jsonapi": {"version": "1.0"}}`

**Screenshots of the issue**
20180608170716-7c011ad592 gif-2-mp4 com

Moto G4 Plus

Yes

Masquerade0097:
Fixes #1038 

Changes: 
1. Created method `getErrorTitleAndDetails` to extract error info under `title` and `detail` tag. Pass this info to user as a snackbar.

Screenshots for the change: 
20180609061133-7c011ad592 gif-2-mp4 com


Masquerade0097:
@iamareebjamal Should I do the same for other HTTP error codes which are hardcoded by strings at the instant ?

iamareebjamal:
Yes, but fallback to default if extraction fails

app/src/main/java/org/fossasia/openevent/app/utils/ErrorUtils.java
```diff
@@ -34,15 +35,16 @@ public static String getMessage(Throwable throwable) {
                 case UNAUTHORIZED:
                     return "Invalid Credentials! Please check your credentials.";
                 case FORBIDDEN:
-                    return "Sorry, you are not authorized to make this request.";
+                    ResponseBody responseBody = ((HttpException) throwable).response().errorBody();
+                    return getErrorTitleAndDetails(responseBody);
```
iamareebjamal: Extract this in a function and only cast if exception is of correct type

Masquerade0097:
@iamareebjamal Is this fine ? Build is failing because tests are not updated. I will update them after finalising code.

app/src/main/java/org/fossasia/openevent/app/utils/ErrorUtils.java
```diff
@@ -75,6 +69,31 @@ public static String getErrorDetails(ResponseBody responseBody) {
         }
     }
 
+    @SuppressWarnings("PMD.AvoidCatchingGenericException")
+    public static String getErrorTitleAndDetails(Throwable throwable) {
```
iamareebjamal: Rather return an object with relevant information set which can be reused for other purposes, then caller can reduce it to String if it wants

mishuvs:
@Masquerade0097 has added around 40 PMD and 10 Checkstyle violations. Please include the fix in your next PR.

PR #1045, opened 2018-06-09
feat: Create Call for Speakers

### LINKED ISSUE #1029: Create Speakers Call
### Feature Request

**Feature Description**

Functionality to create Speakers Call.

**Parent Issue**
#731 
   
Yes

Masquerade0097:
Fixes #1029 

Changes: 
1. Organizer can create speakers call for event if it does not already exist.
2. Database change listener reflects result in UI without explicitly refreshing.

Screenshots for the change: 
20180609143518-7c011ad592 gif-2-mp4 com


iamareebjamal:
Ask to server guys about hash and privacy

Masquerade0097:
@iamareebjamal This is the reply from Mayank about hash and privacy fields -

Hash is used for accessing private call-for-speakers announcement. It is assumed that if you know the hash then you can access the private cfs. It is shown to the organiser during event creation.
Privacy is again if the cfs is private or public
Public cfs is displayed on the frontend in the event page, but private isn't. To access the private one you need the hash.

iamareebjamal:
OK, so they need to be set in Speaker Call?

Masquerade0097:
@iamareebjamal The create speaker call fragment is the only place in our app where organizer can set these.

PR #1096, opened 2018-06-17
feat: Add spinner for session state

### LINKED ISSUE #1085: Bug while Creating Sessions
### Bug Report    

**Actual Behaviour**
While sessions are being created and an alphabet is being entered in **Level**, the app is crashing. 

More over in the **State** edit text there should be a spinner which should consist of all the possible states as it is difficult to comprehend for the organizer the exact keyword which needs to be entered here:
So the possible states that would be included in the spinner will be : Pending, Accepted, Confirmed, Rejected and Draft.

Lollipop 5.1.1 and Redmi Note 3

Yes

sridharjajoo:
Fixes #1085 
Checklist:

- [ ] My branch is up-to-date with the Upstream development branch.

Changes:
- The user can now enter only numbers in the **Level** edit text.
- Added a spinner for selecting state of the session.
 
Screenshots for the change:
ezgif com-resize 16

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/app/core/session/create/CreateSessionPresenter.java
+++ b/app/src/main/java/org/fossasia/openevent/app/core/session/create/CreateSessionPresenter.java
@@ -12,6 +12,8 @@ import org.threeten.bp.LocalDateTime;
 import org.threeten.bp.ZonedDateTime;
 import org.threeten.bp.format.DateTimeParseException;
 
+import java.util.ArrayList;
+import java.util.List;
 import javax.inject.Inject;
 
 import static org.fossasia.openevent.app.common.rx.ViewTransformers.dispose;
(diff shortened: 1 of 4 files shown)


sridharjajoo:
@iamareebjamal When I am updating any particular session, the spinner doesn't show the previously saved state. This also happens in case of editing events i.e If I select any particular time zone from the spinner and update it, it gets updated but when it is again revisited it has the original value.

sridharjajoo:
@iamareebjamal @mishuvs @Masquerade0097   Kindly review

app/src/main/java/org/fossasia/openevent/app/core/session/create/CreateSessionPresenter.java
```diff
@@ -122,4 +125,12 @@ public void createSession(long trackId, long eventId) {
                 getView().dismiss();
             }, Logger::logError);
     }
+
+    public List<String> getSessionStateList() {
+        sessionStateList.add("draft");
+        sessionStateList.add("accepted");
+        sessionStateList.add("pending");
+        sessionStateList.add("confirmed");
+        return sessionStateList;
```
srv-twry: You can add this directly in the resources since the choices are predetermined. More info here: https://developer.android.com/guide/topics/ui/controls/spinner

sridharjajoo:
@srv-twry I have done the required changes.

PR #1120, opened 2018-06-22
feat: Improve login functionality

### LINKED ISSUE #1095: Improving the Login Functionality
### Feature Request

**Feature Description**
When the user logs out of the orga app, he is redirected to the login screen and the login credentials are empty. IMO the login credentials should be autofilled by the user name and the password of the last logged out user.
Eventbrite uses a similar functionality as seen below.
ezgif com-resize 15

Yes.

sridharjajoo:
Fixes #1095 

Checklist:

Changes: 

An EncryptionUtils.class was added which handles the encryption and decryption of user email and password. The keys for the encryption of the sensitive information are obtained from the Android Keystore.
The process of encryption is happening in the following manner:

1. The string to be encrypted is passes to the encryptString( ) of EncryptionUtils from the LoginFragment.
2. Here the keys from the android key-store are obtained and with the help of which the string is encrypted. 
3. When decryption needs to be handled, the keys from the key store are obtained which were used to encrypt the string and the encrypted string is decrypted. 

Screenshots for the change:
ezgif com-resize 23


sridharjajoo:
@iamareebjamal Should I move the methods from the View to the ViewModel?Beacuse if I do so then I will not be able to provide the required context.

iamareebjamal:
See how things are wrapped in other places. The code will go in a separate model and not a view model

For example, see AuthService and AuthHolder in Open Event Android

sridharjajoo:
@iamareebjamal I have moved the method to the AuthServiceImpl. Is this fine?

iamareebjamal:
Logic is still in activities. I gave an example of AuthService. Didn't mean transfer logic there

A new model for encryption will be created and used inside ViewModels and Presenters

Activities and fragments should remain unaware of its existence

app/src/main/java/org/fossasia/openevent/app/data/auth/AuthServiceImpl.java
```diff
@@ -25,6 +29,9 @@
     private final Repository repository;
     private final AuthApi authApi;
 
+    @Inject
+    Context context;
```
srv-twry: Are you sure about this ? Context in a non android class is usually not recommended although this one is the application context so no memory leak I guess.

iamareebjamal: This is Android Context, so no memory leak, but this will not be used in AuthService, but in a separate model

We can use it to wrap APIs used in other models

Like Preference

https://github.com/fossasia/open-event-android/blob/development/app/src/main/java/org/fossasia/openevent/general/data/Preference.kt

Used here

https://github.com/fossasia/open-event-android/blob/development/app/src/main/java/org/fossasia/openevent/general/auth/AuthHolder.kt


Now I can use and mock preference outside of Android Classes, use it as an interface, change implementation and do as I please

Similar to what we are doing in this project in ContextUtils

Another example https://github.com/fossasia/open-event-android/blob/development/app/src/main/java/org/fossasia/openevent/general/data/Resource.kt

app/src/main/java/org/fossasia/openevent/app/data/auth/EncryptionService.java
```diff
@@ -0,0 +1,12 @@
+package org.fossasia.openevent.app.data.auth;
+
+import org.fossasia.openevent.app.data.auth.model.Login;
+
+public interface EncryptionService {
```
iamareebjamal: Encryption Service would do encryption instead of getting email and password

iamareebjamal: In presenter/VM

You still haven't created a model for encryption

sridharjajoo: Oh Sorry. I thought that using the Login model will do as the new encryption model will have the same instance variables (email and password) as the login model

iamareebjamal: The encryption service will not need context at all in its functions. It will be an interface, its implementation will encapsulate context. As mentioned earlier, check preference class mentioned above in conversation with @srv-twry

iamareebjamal: @srv-twry If you were able to get what I was trying to say, can you elaborate the idea and collaborate with @sridharjajoo

sridharjajoo: @iamareebjamal I get your point.

app/src/main/java/org/fossasia/openevent/app/data/auth/model/Encryption.java
```diff
@@ -0,0 +1,21 @@
...
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+public class Encryption {
+
+    @Inject
+    Context context;
```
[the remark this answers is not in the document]
srv-twry: @iamareebjamal Did you mean adding a class like this when you said Encryption needs a separate model or you meants something like a controller like a repository to handle everything for encryption ?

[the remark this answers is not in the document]
iamareebjamal: Exactly @srv-twry Model in MVP does not mean a POJO. It is a repository or service

app/src/main/java/org/fossasia/openevent/app/core/auth/login/LoginViewModel.java
```diff
@@ -96,4 +110,25 @@ protected void onCleared() {
...
+
+    public String getUserpassword() {
+        String password;
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN_MR2) {
+            password = EncryptionUtils.decryptString(encryption.getContext(), sharedPreferenceModel.getString(Constants.PREF_USER_PASSWORD, null));
+        } else {
+            password = sharedPreferenceModel.getString(Constants.PREF_USER_PASSWORD, null);
+        }
+        return password;
+    }
```
srv-twry: What i would suggest doing is create a repository called `EncryptionRepository`. You can move these two methods there and just call it from the vm if needed. This would mean that the VM is like a clean API and the Model does all the heavy lifting.

srv-twry:
Basically your VM only knows about the `EncryptionRepository` and calls methods on it.
The Repository coordinates between the Service and the Utils that you've created. 

@iamareebjamal Is it the required behaviour ?

iamareebjamal:
Repository deals with different sources of data. We'll save EncryptionService

There will be no Utils at all. There'll be an EncryptionService interface

```Java
public interface EncryptionService {

    String encrypt(key, raw);

    String decrypt(key, cypher);

}
```

This kind of thing, and the implementation will contain the encryption logic

```Java

public class AndroidEncryptionService implements EncryptionService {

    @Inject
    public AndroidEncryptionService(Context context) {
        // Initialize whatever here and store in final fields
        // Change implementation according to API levels or whatever
    }
    
    @Override
    public String encrypt(String key, String raw) {
        // Do the heavy lifting
    }
   
    ...

}
```

app/src/main/java/org/fossasia/openevent/app/data/auth/EncryptionServiceImpl.java
```diff
@@ -1,29 +1,36 @@
...
-    public EncryptionServiceImpl(AuthHolder authHolder) {
-        this.authHolder = authHolder;
+    public EncryptionServiceImpl(Context context) {
+        this.context = context;
     }
 
     @Override
-    public void setEncryption(Encryption encryption) {
+    public String encrypt(String credential) {
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN_MR2) {
```
iamareebjamal: We are targetting above Jelly Bean, so why this check?

Also, move the encryption utils code here

app/src/main/java/org/fossasia/openevent/app/data/auth/EncryptionService.java
```diff
@@ -0,0 +1,10 @@
+package org.fossasia.openevent.app.data.auth;
```
iamareebjamal: Encryption is not related to auth

app/src/main/java/org/fossasia/openevent/app/data/auth/EncryptionServiceImpl.java
```diff
@@ -0,0 +1,215 @@
...
+import javax.crypto.CipherInputStream;
+import javax.crypto.CipherOutputStream;
+import javax.inject.Inject;
+import javax.security.auth.x500.X500Principal;
+
+import timber.log.Timber;
+
+public class EncryptionServiceImpl implements EncryptionService {
+
+    private Context context;
```
iamareebjamal: final

app/src/main/java/org/fossasia/openevent/app/data/auth/EncryptionServiceImpl.java
```diff
@@ -0,0 +1,215 @@
...
+    private static final String KEYSTORE = "AndroidKeyStore";
+    private static final String ALIAS = "MY_APP";
```
iamareebjamal: A little better than that

app/src/main/java/org/fossasia/openevent/app/data/auth/EncryptionServiceImpl.java
```diff
@@ -0,0 +1,215 @@
...
+    private static final String ENCODING = "UTF-8";
+
+    @Inject
+    public EncryptionServiceImpl(Context context) {
+        this.context = context;
+    }
+
+    @Override
+    public String encrypt(String credential) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN_MR2) {
```
iamareebjamal: We are targetting above JB, so why this check?

app/src/main/java/org/fossasia/openevent/app/data/auth/EncryptionServiceImpl.java
```diff
@@ -0,0 +1,215 @@
...
+
+        return (KeyStore.PrivateKeyEntry) entry;
+    }
+
+    /**
+     * Creates a public and private key and stores it using the Android Key Store, so that only
+     * this application will be able to access the keys.
+     */
+    @RequiresApi(api = Build.VERSION_CODES.JELLY_BEAN_MR2)
+    private void createKeys(Context context) throws NoSuchProviderException,
```
iamareebjamal: No need to parametrize context

app/src/main/java/org/fossasia/openevent/app/common/Constants.java
```diff
@@ -21,6 +21,8 @@
     public static final String PREF_PAYMENT_ACCEPT_ONSITE = "accept_onsite";
     public static final String PREF_PAYMENT_ONSITE_DETAILS = "onsite_details";
     public static final String PREF_PAYMENT_COUNTRY = "key";
+    public static final String PREF_USER_PASSWORD = "user_password";
+    public static final String PREF_USER_EMAIL = "user_email";
```
iamareebjamal: Move to VM

app/src/main/java/org/fossasia/openevent/app/core/auth/login/LoginFragment.java
```diff
@@ -70,13 +69,24 @@ public void onStart() {
             if (!validator.validate())
                 return;
 
+            loginFragmentViewModel.getLogin().setEmail(binding.emailDropdown.getText().toString());
+            loginFragmentViewModel.getLogin().setPassword(binding.userPassword.getText().toString());
```
iamareebjamal: This shouldn't be needed as you are already getting the email and password in login method

app/src/main/java/org/fossasia/openevent/app/core/auth/login/LoginFragment.java
```diff
@@ -70,13 +69,24 @@ public void onStart() {
...
+
             ViewUtils.hideKeyboard(view);
             loginFragmentViewModel.login();
+            loginFragmentViewModel.encryption();
```
iamareebjamal: It's the responsibility of VM to do whatever is required and not rely on view to tell what to do, this call is never done without login call, so should be implicitly coupled with it

app/src/main/java/org/fossasia/openevent/app/core/auth/login/LoginFragment.java
```diff
@@ -70,13 +69,24 @@ public void onStart() {
...
         });
         binding.signUpLink.setOnClickListener(view -> openSignUpPage());
         binding.forgotPasswordLink.setOnClickListener(view -> openForgotPasswordPage());
     }
 
+    @Override
+    public void onResume() {
+        super.onResume();
+        binding.emailDropdown.setText(loginFragmentViewModel.getDecryptedEmail());
+        binding.userPassword.setText(loginFragmentViewModel.getDecryptedPassword());
```
iamareebjamal: Again, VMs responsibility to send this to view, not view's responsibility to pull

app/src/main/java/org/fossasia/openevent/app/data/auth/EncryptionServiceImpl.java
sridharjajoo: We have a min API level of 16 and this requires API Level 18

app/src/main/java/org/fossasia/openevent/app/core/auth/login/LoginFragment.java
sridharjajoo: This is needed because I have removed the setting of password and email ("@={login.email}" and "@={login.password}") from the XML file as the saved value in shared preference wasn't been shown in the edit text. So I am setting the values programmatically on click.

sridharjajoo:
@iamareebjamal I had already created a EncryptionServiceImpl (model class) in the second commit itself.([link](https://github.com/fossasia/open-event-orga-app/pull/1120/commits/11002cc11488528ab02aa1a6debd72759d7b397b)). So was not able to understand the model class which you were talking about earlier.

app/src/main/java/org/fossasia/openevent/app/data/auth/EncryptionServiceImpl.java
iamareebjamal: So, on versions below 18, it will not encrypt the String, this is a very big security issue

app/src/main/java/org/fossasia/openevent/app/core/auth/login/LoginFragment.java
iamareebjamal: I think if you let the previous method remain and bind the saved value correctly to View, it will work

app/src/main/java/org/fossasia/openevent/app/data/auth/EncryptionServiceImpl.java
sridharjajoo: The **KeyStoreGeneratorSpec** which is used to generate and stores it using android keystore was introduced in API level 18 only. To handle cases before that we can encrypt it normally without having the need to use KEYS. Should I use that for this case?

app/src/main/java/org/fossasia/openevent/app/core/auth/login/LoginFragment.java
sridharjajoo: Should livedata be used here?

app/src/main/java/org/fossasia/openevent/app/data/auth/EncryptionServiceImpl.java
iamareebjamal: Yes

app/src/main/java/org/fossasia/openevent/app/data/encryption/EncryptionServiceImpl.java
```diff
@@ -60,16 +60,16 @@ public String encrypt(String credential) {
...
-            return decryptString(context, encryptedCredentials);
+            return decryptString(encryptedCredentials);
         } else {
             return encryptedCredentials;
         }
     }
 
     @RequiresApi(api = Build.VERSION_CODES.JELLY_BEAN_MR2)
-    private String encryptString(Context context, String toEncrypt) {
+    private String encryptString(String toEncrypt) {
```
iamareebjamal: Handle nullability here

app/src/main/java/org/fossasia/openevent/app/core/auth/login/LoginFragment.java
iamareebjamal: Still there and even two extra endpoints are added. We need to decrease calls in views, not increase

app/src/main/java/org/fossasia/openevent/app/core/auth/login/LoginFragment.java
```diff
@@ -64,6 +63,8 @@ public void onStart() {
         loginFragmentViewModel.getProgress().observe(this, this::showProgress);
         loginFragmentViewModel.getError().observe(this, this::showError);
         loginFragmentViewModel.getEmailList().observe(this, this::attachEmails);
+        loginFragmentViewModel.getDecryptedEmail().observe(this, email -> binding.emailDropdown.setText(email));
+        loginFragmentViewModel.getDecryptedPassword().observe(this, password -> binding.userPassword.setText(password));
```
iamareebjamal: Even two extra endpoints are added. We need to decrease calls in views, not increase

sridharjajoo: I haven't removed them  because the edit text fields aren't getting set if I am setting them via xml. I have tested it. I can upload the gif as well.

iamareebjamal: Did you push the loaded encrypted value through the LiveData, because I haven't seen that

Ideally, you should be pushing the login and binding it. The two way binding will also work then

sridharjajoo: Why would I need to push the encrypted value through LiveData. I just want to observe the decrypted string in the view

app/src/main/java/org/fossasia/openevent/app/core/auth/login/LoginFragment.java
```diff
@@ -64,14 +63,20 @@ public void onStart() {
         loginFragmentViewModel.getProgress().observe(this, this::showProgress);
         loginFragmentViewModel.getError().observe(this, this::showError);
         loginFragmentViewModel.getEmailList().observe(this, this::attachEmails);
+
+        loginFragmentViewModel.getLogin().observe(this, login -> {
+            binding.emailDropdown.setText(login.getEmail());
+            binding.userPassword.setText(login.getPassword());
```
iamareebjamal: Why not set binding and enable 2 way binding?

app/src/main/java/org/fossasia/openevent/app/core/auth/login/LoginFragment.java
```diff
@@ -64,14 +63,20 @@ public void onStart() {
...
+      
         loginFragmentViewModel.getIsTokenSent().observe(this, (value) -> {
             getFragmentManager().beginTransaction()
                 .setCustomAnimations(R.anim.enter_from_left, R.anim.exit_from_right)
                 .replace(R.id.fragment_container, new ResetPasswordFragment())
                 .commit();
         });
 
-        binding.setLogin(loginFragmentViewModel.getLogin());
+        binding.setLogin(loginFragmentViewModel.getLogin().getValue());
```
iamareebjamal: There'll be no need of this once the above observer is fixed

app/src/main/java/org/fossasia/openevent/app/core/auth/login/LoginViewModel.java
```diff
@@ -76,8 +91,13 @@ public void setBaseUrl(String url, boolean shouldSetDefaultUrl) {
...
     }
 
-    public Login getLogin() {
-        return login;
+    public LiveData<Login> getLogin() {
+        String email = encryptionService.decrypt(sharedPreferenceModel.getString(PREF_USER_EMAIL, null));
+        String password = encryptionService.decrypt(sharedPreferenceModel.getString(PREF_USER_PASSWORD, null));
+        login.setEmail(email);
+        login.setPassword(password);
+        decryptedLogin.setValue(login);
```
iamareebjamal: Only initialize once

app/src/main/java/org/fossasia/openevent/app/data/encryption/EncryptionServiceImpl.java
```diff
@@ -0,0 +1,286 @@
...
+             e.printStackTrace();
+         } catch (InvalidKeyException e) {
+             e.printStackTrace();
+         } catch (IllegalBlockSizeException e) {
+             e.printStackTrace();
+         } catch (BadPaddingException e) {
+             e.printStackTrace();
+         } catch (UnsupportedEncodingException e) {
+             e.printStackTrace();
+         }
```
iamareebjamal: Use Timber and collapse the catch blocks

app/src/main/java/org/fossasia/openevent/app/core/auth/login/LoginViewModel.java
```diff
@@ -99,6 +119,11 @@ protected void onCleared() {
         compositeDisposable.dispose();
     }
 
+    //used only for testing
+    public Login getDecryptedLoginCredentials() {
```
iamareebjamal: Add `@VisibleForTesting`

app/src/main/java/org/fossasia/openevent/app/core/auth/login/LoginViewModel.java
```diff
@@ -24,33 +26,47 @@
...
 
     //for logging into the app
     public void login() {
         compositeDisposable.add(loginModel.login(login)
             .doOnSubscribe(disposable -> progress.setValue(true))
             .doFinally(() -> progress.setValue(false))
-            .subscribe(() -> isLoggedIn.setValue(true),
+            .subscribe(() -> {  isLoggedIn.setValue(true);
+                                encryptUserCredentials();
+                             },
```
iamareebjamal: Everything seems good. Just properly format this line and encrypt before sending true

app/src/main/java/org/fossasia/openevent/app/core/auth/login/LoginViewModel.java
```diff
@@ -24,33 +26,47 @@
...
                 throwable -> error.setValue(ErrorUtils.getMessage(throwable).toString())));
     }
-
+
```
iamareebjamal: Extra whitespace

app/src/main/java/org/fossasia/openevent/app/core/auth/login/LoginViewModel.java
```diff
@@ -76,8 +92,11 @@ public void setBaseUrl(String url, boolean shouldSetDefaultUrl) {
         interceptor.setInterceptor(baseUrl);
     }
 
-    public Login getLogin() {
-        return login;
+    public LiveData<Login> getLogin() {
+        login.setEmail(encryptionService.decrypt(sharedPreferenceModel.getString(PREF_USER_EMAIL, null)));
+        login.setPassword(encryptionService.decrypt(sharedPreferenceModel.getString(PREF_USER_PASSWORD, null)));
+        decryptedLogin.setValue(login);
```
iamareebjamal: Only do this once

app/src/main/res/layout/login_fragment.xml
```diff
@@ -51,7 +51,7 @@
                         android:layout_width="match_parent"
                         android:layout_height="wrap_content"
                         android:hint="@string/email"
-                        android:text="@={ login.email }"
+                        android:text="@={login.email}"
```
iamareebjamal: Revert to original

app/src/main/res/layout/login_fragment.xml
```diff
@@ -64,10 +64,11 @@
                     android:paddingBottom="@dimen/spacing_normal">
 
                     <android.support.design.widget.TextInputEditText
+                        android:id="@+id/user_password"
```
iamareebjamal: Revert to original

app/src/main/res/layout/login_fragment.xml
```diff
@@ -64,10 +64,11 @@
...
                         android:layout_width="match_parent"
                         android:layout_height="wrap_content"
                         android:hint="@string/password"
-                        android:text="@={ login.password }"
+                        android:text="@={login.password}"
```
iamareebjamal: Revert to original

app/src/main/java/org/fossasia/openevent/app/data/encryption/EncryptionServiceImpl.java
```diff
@@ -0,0 +1,266 @@
...
+        final KeyPair kp = kpGenerator.generateKeyPair();
+        Timber.d("Public Key is: " + kp.getPublic().toString());
+    }
+
+    /*
+    * The following methods are specifically taken into use when the API level is
+    * less than 18
+    *
+    */
+     private byte[] encryptMsg(String message, SecretKey secret) {
```
iamareebjamal: encryptString

app/src/main/java/org/fossasia/openevent/app/data/encryption/EncryptionServiceImpl.java
```diff
@@ -0,0 +1,266 @@
...
+             cipherText = cipher.doFinal(message.getBytes("UTF-8"));
+         } catch (NoSuchAlgorithmException  | NoSuchPaddingException | InvalidKeyException | IllegalBlockSizeException | BadPaddingException | UnsupportedEncodingException e) {
+            Timber.w(e);
+         }
+         return cipherText;
+     }
+        return null;
+     }
+
+    private String decryptMsg(byte[] cipherText, SecretKey secret) {
```
iamareebjamal: decryptString

app/src/main/java/org/fossasia/openevent/app/data/encryption/EncryptionServiceImpl.java
```diff
@@ -0,0 +1,266 @@
...
+     if (message != null) {
+         try {
+             cipher = Cipher.getInstance("AES/ECB/PKCS5Padding");
+             cipher.init(Cipher.ENCRYPT_MODE, secret);
+             cipherText = cipher.doFinal(message.getBytes("UTF-8"));
+         } catch (NoSuchAlgorithmException  | NoSuchPaddingException | InvalidKeyException | IllegalBlockSizeException | BadPaddingException | UnsupportedEncodingException e) {
+            Timber.w(e);
+         }
+         return cipherText;
+     }
```
iamareebjamal: Wrong indentation

app/src/main/java/org/fossasia/openevent/app/data/encryption/EncryptionServiceImpl.java
```diff
@@ -0,0 +1,266 @@
...
+    private void createKeys() throws NoSuchProviderException,
+        NoSuchAlgorithmException, InvalidAlgorithmParameterException {
+        // Create a start and end time, for the validity range of the key pair that's about to be
+        // generated.
+        Calendar start = new GregorianCalendar();
+        Calendar end = new GregorianCalendar();
+        end.add(Calendar.YEAR, 25);
+
+        KeyPairGeneratorSpec spec =
+            new KeyPairGeneratorSpec.Builder(context)
```
iamareebjamal: Move to previous line

app/src/main/java/org/fossasia/openevent/app/core/auth/login/LoginViewModel.java
iamareebjamal: The setting of email, password, decrypting of email, password, and setting of decrypted login

sridharjajoo: I am already doing it once.

app/src/main/java/org/fossasia/openevent/app/core/auth/login/LoginViewModel.java
```diff
@@ -24,30 +26,45 @@
...
+    private final String PREF_USER_EMAIL = "user_email";
 
     private final MutableLiveData<Boolean> progress = new MutableLiveData<>();
     private final MutableLiveData<String> error = new MutableLiveData<>();
+    private final MutableLiveData<Login>  decryptedLogin = new MutableLiveData<>();
     private MutableLiveData<Boolean> isLoggedIn;
     private final MutableLiveData<Boolean> isTokenSent = new MutableLiveData<>();
     private MutableLiveData<Set<String>> emailList;
-
+
```
iamareebjamal: Extra whitespace

app/src/main/java/org/fossasia/openevent/app/core/auth/login/LoginViewModel.java
```diff
@@ -99,7 +116,12 @@ protected void onCleared() {
         compositeDisposable.dispose();
     }
 
-    public void requestToken() {
+    @VisibleForTesting
+    public Login getLoginModel() {
+        return login;
+    }
+
+  public void requestToken() {
```
iamareebjamal: Wrong indentation

app/src/main/java/org/fossasia/openevent/app/core/auth/login/LoginViewModel.java
```diff
@@ -76,8 +93,8 @@ public void setBaseUrl(String url, boolean shouldSetDefaultUrl) {
         interceptor.setInterceptor(baseUrl);
     }
 
-    public Login getLogin() {
-        return login;
+    public LiveData<Login> getLogin() {
+        return decryptedLogin;
```
iamareebjamal: You removed all the logic, where are you loading the decrypted data?

app/src/main/java/org/fossasia/openevent/app/data/encryption/EncryptionServiceImpl.java
```diff
@@ -0,0 +1,265 @@
...
+        Timber.d("Public Key is: " + kp.getPublic().toString());
+    }
+
+    /*
+    * The following methods are specifically taken into use when the API level is
+    * less than 18
+    *
+    */
+    private byte[] encryptString(String message, SecretKey secret) {
+
```
iamareebjamal: Extra line

app/src/main/java/org/fossasia/openevent/app/core/auth/login/LoginViewModel.java
sridharjajoo: I pushed by mistake

app/src/main/java/org/fossasia/openevent/app/core/auth/login/LoginViewModel.java
iamareebjamal: Every time getLogin is called, this happens

Should only happen once

app/src/main/java/org/fossasia/openevent/app/core/auth/login/LoginViewModel.java
```diff
@@ -76,8 +93,14 @@ public void setBaseUrl(String url, boolean shouldSetDefaultUrl) {
...
-    public Login getLogin() {
-        return login;
+    public LiveData<Login> getLogin() {
+        if (decryptedLogin.getValue() == null) {
+            login.setEmail(encryptionService.decrypt(sharedPreferenceModel.getString(PREF_USER_EMAIL, null)));
+            login.setPassword(encryptionService.decrypt(sharedPreferenceModel.getString(PREF_USER_PASSWORD, null)));
+            decryptedLogin.setValue(login);
+            return decryptedLogin;
+        }
+        return null;
```
iamareebjamal: NPE

PR #1125, opened 2018-06-23
chore: remove forgot password screen; reset password module to MVVM

### LINKED ISSUE #1100: Directly send token when clicked on Forgot Password
**Is it a ?**

- [x] Chore
    
**Chore Description**

Clicking Forgot Password opens another screen which requires clicking 'Request Token'. 
It should directly send the token.

Yes

mishuvs:
Fixes #1100 

Checklist:

Changes: 
1. Removes the forgot password screen
2. Removes package auth.fogot.request
3. Renames package auth.forgot.submit to auth.reset
4. Improves Reset Password screen's layout
5. Converts auth.reset module to MVVM

GIF for the change:
videotogif_2018 06 23_22 45 31


srv-twry:
@mishuvs Please attach a GIF of the current procedure of password reset.

app/src/main/java/org/fossasia/openevent/app/core/auth/login/LoginFragment.java
```diff
@@ -64,6 +64,14 @@ public void onStart() {
         loginFragmentViewModel.getProgress().observe(this, this::showProgress);
         loginFragmentViewModel.getError().observe(this, this::showError);
         loginFragmentViewModel.getEmailList().observe(this, this::attachEmails);
+        loginFragmentViewModel.getIsTokenSent().observe(this, (value) -> {
+            if (value) {
```
srv-twry: We usually try to keep the view logic free. What is the need of this if statement ? It seems like this can be done using a `SingleEventLiveData` that I added yesterday. Please check it out if it serves your purpose also.

mishuvs:
Yes, wait :sweat_smile: .. don't generally get a review that fast
@srv-twry Please review again.

app/src/main/java/org/fossasia/openevent/app/core/auth/login/LoginViewModel.java
```diff
@@ -96,4 +98,19 @@ protected void onCleared() {
         super.onCleared();
         compositeDisposable.dispose();
     }
+
+    public void requestToken() {
+        RequestToken requestToken = new RequestToken();
+        requestToken.setEmail(login.getEmail());
+
+        compositeDisposable.add(loginModel.requestToken(requestToken)
```
srv-twry: Instead of adding this, use
`.compose(dispose(compositeDisposable))`. We already have it defined in the `ViewTransformers` file.

app/src/main/java/org/fossasia/openevent/app/core/auth/reset/ResetPasswordViewModel.java
```diff
@@ -0,0 +1,82 @@
...
+        this.tokenSubmitModel = tokenSubmitModel;
+        this.interceptor = interceptor;
+    }
+
+    public SubmitToken getSubmitToken() {
+        return submitToken;
+    }
+
+    public void submitRequest() {
+        compositeDisposable.add(tokenSubmitModel.submitToken(submitToken)
```
srv-twry: Same here.

app/src/main/java/org/fossasia/openevent/app/core/auth/login/LoginFragment.java
sridharjajoo: Yeah, Try to send the value to the view only when the return value from getIsTokenSent() is true. Then you'll not need to use the conditional statement here.

mishuvs: @srv-twry Thanks for pointing out, the check can be removed as the fragment is recreated every time.

mishuvs:
Made the change @srv-twry @Masquerade0097 @sridharjajoo @iamareebjamal 
Please review.

app/src/main/java/org/fossasia/openevent/app/core/auth/login/LoginFragment.java
```diff
@@ -64,6 +64,12 @@ public void onStart() {
         loginFragmentViewModel.getProgress().observe(this, this::showProgress);
         loginFragmentViewModel.getError().observe(this, this::showError);
         loginFragmentViewModel.getEmailList().observe(this, this::attachEmails);
+        loginFragmentViewModel.getIsTokenSent().observe(this, (value) -> {
+            getFragmentManager().beginTransaction()
+                .setCustomAnimations(R.anim.enter_from_left, R.anim.exit_from_right)
+                .replace(R.id.fragment_container, new ResetPasswordFragment())
+                .commit();
```
srv-twry: What this would do is create a new `ResetPasswordFragment` everytime you rotate the screen and the `value` is set to true.

srv-twry: Even if it doesn't then also it's not good for readability. A `SingleActionLiveData` has been added specifically for this purpose. Check how we are opening dashboard/event list in the `MainActivity` and the `EventViewModel` of the `main` module.

mishuvs: > What this would do is create a new ResetPasswordFragment everytime you rotate the screen and the value is set to true.

Didn't understand this. Once the value is set, ResetPasswordFragment will be replaced.
> Even if it doesn't then also it's not good for readability

Not sure how using a custom class improves readability.
> A `SingleActionLiveData` has been added specifically for this purpose.

`SingleEventLiveData` I think is being used for events which can be triggered repeatedly, but should notify only once. But in this case, the 'I forgot my password' button can be clicked only once in the lifecycle. Once it's clicked, the fragment will be destroyed.
`SingleEventLiveData` seemed to be right in case of adding fragment instead of replacing, but even that is not true I think, because then the `SingleEventLiveData` won't send be able to send anymore notifications, which is crucial for continual functioning of the 'I forgot my password' button even after fragment recreation.

iamareebjamal: That's not what SingleEventLiveData is used for @mishuvs 

It just doesn't retransmit the last value on rotation. That's it

mishuvs: Okay @iamareebjamal , but what I am not able to understand is what does it has to do with rotation?
The value is false all the time -> Button clicked -> value sets to  true -> ResetPasswordFragment replaced and eventually destroyed -> if back to login, the value again starts with false
Should I anyway use SingleEventLiveData?

app/src/main/res/layout/reset_password_by_token_fragment.xml
```diff
@@ -29,6 +29,23 @@
...
+                    android:layout_gravity="center_horizontal"
+                    android:textColor="@color/color_accent"
+                    android:textSize="@dimen/text_size_larger"
+                    style="@style/Base.TextAppearance.AppCompat.Medium"
+                    android:layout_marginBottom="@dimen/spacing_normal"/>
+
+                <TextView
+                    android:layout_width="wrap_content"
+                    android:layout_height="wrap_content"
+                    android:text="@string/check_email_reset_password_msg"
```
Masquerade0097: It appears that this string is displayed even when there is an error in token request and token is not sent to email which should not be the case.

PR #1133, opened 2018-06-27
feat: Add suggestions in navigation drawer

### LINKED ISSUE #1108: use feedback function
developers,
use feedback function because then you can get idea from feedback and then you can develop this app.

sridharjajoo:
Fixes #1108 

Checklist:

Changes: A Google form was created its shareable link was added in the Intent.

Screenshots for the change:
ezgif com-resize 25

### DIFF
--- a/app/src/main/java/org/fossasia/openevent/app/core/main/DrawerNavigator.java
+++ b/app/src/main/java/org/fossasia/openevent/app/core/main/DrawerNavigator.java
@@ -2,6 +2,7 @@ package org.fossasia.openevent.app.core.main;
 
 import android.content.Context;
 import android.content.Intent;
+import android.net.Uri;
 import android.support.annotation.VisibleForTesting;
 import android.support.v7.app.AlertDialog;
 import android.view.MenuItem;
@@ -19,6 +20,7 @@ class DrawerNavigator {
     private final FragmentNavigator fragmentNavigator;
 
     private AlertDialog logoutDialog;
+    private final String GOOGLE_FORM_LINK = "https://docs.google.com/forms/d/e/1FAIpQLSfJ-v1mbmNp1ChpsikHDx6HZ5G9Bq8ELCivckPPcYlOAFOy2Q/viewform?usp=sf_link";
 
     DrawerNavigator(Context context, FragmentNavigator fragmentNavigator, MainPresenter mainPresenter) {
         this.context = context;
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/fossasia/openevent/app/core/main/DrawerNavigator.java
```diff
@@ -19,6 +20,7 @@
     private final FragmentNavigator fragmentNavigator;
 
     private AlertDialog logoutDialog;
+    private final String GOOGLE_FORM_LINK= "https://docs.google.com/forms/d/e/1FAIpQLSfJ-v1mbmNp1ChpsikHDx6HZ5G9Bq8ELCivckPPcYlOAFOy2Q/viewform?usp=sf_link";
```
iamareebjamal: Where does this Google form submit

sridharjajoo: orga.developers@gmail.com

iamareebjamal: Space before =

sridharjajoo:
@iamareebjamal Any more changes?

app/src/main/java/org/fossasia/openevent/app/core/main/DrawerNavigator.java
```diff
@@ -19,6 +20,7 @@
     private final FragmentNavigator fragmentNavigator;
 
     private AlertDialog logoutDialog;
+    private final String GOOGLE_FORM_LINK = "https://docs.google.com/forms/d/e/1FAIpQLSfJ-v1mbmNp1ChpsikHDx6HZ5G9Bq8ELCivckPPcYlOAFOy2Q/viewform?usp=sf_link";
```
srv-twry: You could probably add it in the Constants class where we store all our constants.

sridharjajoo: I added it here because its being used in this class only. So it would be better to declare it here rather than using a Constants class for it.

srv-twry: Ok, works in both ways.

iamareebjamal: Yes, even constants should be used as locally as possible. We should not pollute global namespace

PR #1179, opened 2018-07-08
fix: Decrease tab elevation

### LINKED ISSUE #1173: Double elevation in tabs and action bar
### Bug Report    

**Actual Behaviour**
Double elevation is present in tabs and action bar of the app which needs to be removed.

Android 6.0.1 and Redmi Note 3

Yes

sridharjajoo:
Fixes #1173 

Checklist:

- [x] I have checked for PMD and check-style issue

Screenshots for the change:
ezgif com-resize 36


sridharjajoo:
@mishuvs @iamareebjamal I think if the elevation of both is made the same then it wouldn't look nice as the toolbar color and tabs color is the same (both white) and hence both would look merged with each other.

sridharjajoo:
@iamareebjamal I have updated the P.R. One thing that is happening is that when the user logs in , the elevation is there in the events list but when an event is selected and then we again move to the events list, then the elevation is gone. I have shown the gif above.

PR #1288, opened 2018-08-02
fix: Add multiline support for Update Event

### LINKED ISSUE #1287: Add MultiLine for UpdateEvent EditText
### Bug Report    

**Actual Behaviour**
As per the meeting notes, currently while filling the update event form, there is no multi line support. It needs to be added for certain fields such as Organizer description, Code of Conduct etc.

Yes

sridharjajoo:
Fixes #1287 

Checklist:


mishuvs:
Where do you think it will be used after Update Event screen is linked to the Steppers?

sridharjajoo:
What do you think should be done? You have created an issue about making the UpdateFragment more similar to Eventbrite (where the stepper isn't used and just a single screen is there)

mishuvs:
We have to keep both create and update screens look the same, as directed in the meeting. No point using duplicate layouts. Since the stepper is similar to EventBrite (I thought it was), I was referring to that. I have now changed the title. You can however make changes to the Stepper layout if needed.

sridharjajoo:
In eventbrite the update event is not a stepper, creation of events is

mishuvs:
Thanks, I just checked.
Actually, even their Create Event is not a stepper if looked like that.
They are using a series of screens only to take the values Name -> Description -> Date
Then they open a Detail Screen, which is equivalent to the Stepper in our case since we are showing a lot more info there. This is where the user enters more details at the time of creation.
This Detail Screen is the same screen they are using in Update Event. 
The point is to use the same screen for both the things. 
Last time Mario asked us to show the same screen in both cases, that's why I insisted on reusing.

sridharjajoo:
This is what mario said:
He said that the fields present in the UpdateEvent screen of Eventbrite and Update Event screen of our app aren't the same i.e the ordering of these fields (location, profile pic, timezone etc.) is different and hence we need to make it more consistent with eventbrite so only the ordering needs to be changed.

Their details screen and our update screen are one and the same thing. If we convert the update screen to stepper then the consistency will be gone and the same issue about consistency between eventbrite and our app will be repeated in the meeting.

Following are the 2 screens. If you convert the edit event in our app to stepper will the consistency still be there?

ezgif com-resize 7
ezgif com-resize 6

If you still want, I can close this P.R

Issue #1402, opened 2018-12-01
Can't Create Event if  Start Time and End time is same "But date is different"


ankush-01:
Current Behaviour :
Can't create event if start and end time is same but date is different.

Expected Behaviour: 
their should be no restriction on event Time like "Event of 24 hours(1 Day) which starts and end at same time"

ShridharGoel:
Not able to reproduce this.

Issue #1400, opened 2018-11-30
Their should be Minimum and maximum duration of Event.


ankush-01:
Current Behaviour:
 user can create 1 minute of event to event having 10-15 (or more) years to finish.
Expected Behaviour:
 their should be minimum and maximum duration of event. 
![1de281d0-15bb-4e26-9ac9-0b4ed2c4cc45](https://user-images.githubusercontent.com/33266550/49315472-a7afdd80-f4a2-11e8-9610-e028d66d6fe4.jpg)

Masquerade0097:
I think it's better if server provides this restriction rather than using some hard coded values here.

iamareebjamal:
I don't think that it is even required honestly. You should be able to create a 5 minute event as well

ankush-01:
> I think it's better if server provides this restriction rather than using some hard coded values here.

no its simple ...it can be implemented only in 5-10 lines of code..

iamareebjamal:
Yeah, they should be able to do so unless server restricts it

PR #1403, opened 2018-12-01
Update CreateEventViewModel.java

ankush-01:
#1402
#1400

### DIFF
--- a/app/src/main/java/com/eventyay/organizer/core/event/create/CreateEventViewModel.java
+++ b/app/src/main/java/com/eventyay/organizer/core/event/create/CreateEventViewModel.java
@@ -105,9 +105,27 @@ public class CreateEventViewModel extends ViewModel {
             ZonedDateTime start = DateUtils.getDate(event.getStartsAt());
             ZonedDateTime end = DateUtils.getDate(event.getEndsAt());
 
-            if (!end.isAfter(start)) {
+            if (!end.isAfter(start) && start.getDayOfYear()==end.getDayOfYear() && start.getYear()==end.getYear()) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/eventyay/organizer/core/event/create/CreateEventViewModel.java
```diff
@@ -105,10 +105,28 @@ public boolean verify() {
...
+                    }
+                }else if(start.getYear()<end.getYear()){
+                    if(end.getDayOfYear()<=2 && start.getDayOfYear()>=364){              // event max limit 3 days
+                        return true;
+                    }
+                }
+                onError.setValue("Event Exceeds Maximum Duration Limit");
+                return false;
+
+            }
```
iamareebjamal: We don't want this logic, sorry

app/src/main/java/com/eventyay/organizer/core/event/create/CreateEventViewModel.java
```diff
@@ -105,10 +105,28 @@ public boolean verify() {
             ZonedDateTime start = DateUtils.getDate(event.getStartsAt());
             ZonedDateTime end = DateUtils.getDate(event.getEndsAt());
 
-            if (!end.isAfter(start)) {
+            if (!end.isAfter(start) && start.getDayOfYear()==end.getDayOfYear() && start.getYear()==end.getYear()) {
```
iamareebjamal: Since this is Zoned**DateTime**, this should work without any extra check, so please check ehy is it not working instead of adding extra checks

PR #1417, opened 2019-01-02
fix: Display error if event creation fails

### LINKED ISSUE #1404: Can't Proceed After Step 3 Of Create New Event .
Current Behaviour :
   can't proceed after step 3rd of create new activity even all field is filled.
   Their is no error "Why event is not created".

Expected Behaviour :
   event is generated when user filled out all details and click on CREATE button.
   Error should be displayed like by  Toast Or TextInputLayout Error.

bug

ShridharGoel:
Fixes #1404  
Error needs to displayed when user can't proceed

Checklist:

Changes: 

Now, error message would appear if user cannot proceed after step 3 of event creation.

### DIFF
--- a/app/src/main/java/com/eventyay/organizer/core/event/create/CreateEventViewModel.java
+++ b/app/src/main/java/com/eventyay/organizer/core/event/create/CreateEventViewModel.java
@@ -30,6 +30,7 @@ import java.util.TimeZone;
 import javax.inject.Inject;
 
 import io.reactivex.disposables.CompositeDisposable;
+import timber.log.Timber;
 
 import static com.eventyay.organizer.common.Constants.PREF_ACCEPT_BANK_TRANSFER;
 import static com.eventyay.organizer.common.Constants.PREF_ACCEPT_CHEQUE;
@@ -147,8 +148,8 @@ public class CreateEventViewModel extends ViewModel {
                 onSuccess.setValue("Event Created Successfully");
                 close.setValue(null);
             }, throwable -> {
-                Logger.logError(throwable);
-                onError.setValue(ErrorUtils.getMessage(throwable).toString());
+                onError.setValue(ErrorUtils.getErrorDetails(throwable).toString());
+                Timber.e(throwable, "An exception occurred : %s", throwable.getMessage());
             }));
     }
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/eventyay/organizer/core/event/create/CreateEventViewModel.java
```diff
@@ -138,7 +138,10 @@ public void createEvent() {
             .subscribe(createdEvent -> {
                 onSuccess.setValue("Event Created Successfully");
                 close.setValue(null);
-            }, Logger::logError));
+            }, throwable -> {
+                Logger.logError(throwable);
+                onError.setValue(throwable.getMessage());
+            }));
```
iamareebjamal: No, handle the error and display proper message

ShridharGoel: Done

app/src/main/java/com/eventyay/organizer/core/event/create/CreateEventViewModel.java
```diff
@@ -138,7 +139,10 @@ public void createEvent() {
             .subscribe(createdEvent -> {
                 onSuccess.setValue("Event Created Successfully");
                 close.setValue(null);
-            }, Logger::logError));
+            }, throwable -> {
+                Logger.logError(throwable);
```
iamareebjamal: Use timber

app/src/main/java/com/eventyay/organizer/core/event/create/CreateEventViewModel.java
iamareebjamal: What about unhandled error? It should not print exceptions for unhandled errors

app/src/main/java/com/eventyay/organizer/core/event/create/CreateEventViewModel.java
ShridharGoel: Done

ShridharGoel: Just for info, why timber here when the logError() function is also doing the same thing?

app/src/main/java/com/eventyay/organizer/core/event/create/CreateEventViewModel.java
ShridharGoel: Updated

iamareebjamal: It should not show toasts for unhandled errors

PR #1550, opened 2019-03-10
fix: Use saveAlertDialog when back home button in action bar

### LINKED ISSUE #1519: saveAlertDialog is not used when the upper back button is used in UpdateOrganizerInfoFragment
**Describe the bug**
saveAlertDialog is not used when the upper back button is used in UpdateOrganizerInfoFragment

**Expected behavior**
saveAlertDialog appears

**Screenshots**

Yes

anhanh11001:
Details:
I don't know why when I try using Action Bar, setDisplayShowHomeEnabled(true), setHomeButtonEnabled(true), setDisplayHomeAsUpEnabled(true),... all of that but when onItemOptionMenuClicked, I called android.R.id.home on R.id.home, the back home button is not triggered.

So that's why I use toolbar and set a navigation icon for it. I also add saveAlertDialog for it so that the saveAlertDialog is not only used when backPressed() but also when clicking on up home button

This resolves issue 1519

Fixes #1519 

Screenshots for the change:


iamareebjamal:
This is a workaround, not a fix.

PR #1551, opened 2019-03-10
fix: DatePicker and TimePicker doesn't display time chosen before when clicked

### LINKED ISSUE #1548: TimePicker, DatePicker doesn't display current time
**Describe the bug**
The time picker in create new event fragment doesn't display the current time after changing the time once. The same goes for Date Picker to the left

**Screenshots**

Yes

anhanh11001:
Fix: DatePicker and TimePicker doesn't display time chosen before when clicked

Fixes #1548 

Changes: 
I change the use of ZonedDateTime by the currentTime chosen before so that whenever we click on TimePicker or DatePicker, it shows the current time and date chosen

Screenshots for the change:


iamareebjamal:
There was a reason we were using ZonedDateTime

iamareebjamal:
@anhanh11001 Please update

PR #1555, opened 2019-03-12
chore: Switch to a different colour picker

### LINKED ISSUE #1520: Migrate to a better colour picker in Tracks fragment
**Describe the chore**
https://github.com/duanhong169/ColorPicker

This library provides a better colour picker.

Open to all.

ShridharGoel:
Fixes #1520 

Checklist:

Changes: 

Now, https://github.com/duanhong169/ColorPicker has been used for the colour picker in Tracks fragment.

GIF for the change:

ezgif com-video-to-gif 49


app/src/main/java/com/eventyay/organizer/core/track/create/CreateTrackFragment.java
```diff
@@ -72,19 +71,25 @@ public void onStart() {
...
 
         binding.form.colorPicker.setOnClickListener(view -> {
-            colorPickerDialog.show();
-        });
-
-        colorPickerDialog.setCallback(color -> {
-            binding.form.trackColor.setText(String.format("#%06X", 0xFFFFFF & color));
-            binding.form.colorPicker.setBackgroundColor(color);
-            colorPickerDialog.dismiss();
+            new ColorPickerPopup.Builder(getContext())
```
iadeelzafar: How about adding this colorPicker listener as a method in a related class and then reusing it in both CreateTrackFragment and UpdateTrackFragment?

ShridharGoel: We can shift it to Utils but I don't think there is a need for that since even earlier it was done like this only.

ShridharGoel:
@iamareebjamal Please review this

ShridharGoel:
The earlier colour picker required manual adjustment of RGB values which might not be good from user's point of view.

iamareebjamal:
It's meant for organizers. But OK, if you want to change it, use a better library which aligns with material design. This doesn't look good enough

mariobehling:
Closing this as we are waiting for a better library.

PR #1673, opened 2019-05-27
fix: Update check-in system to work properly

### LINKED ISSUE #1672: Check-in mechanism not working properly
**Describe the bug**
Attendee IDs are 0 due to which the check-in system is not functioning properly.

**GIF**
ezgif com-video-to-gif (78)

- [x] Yes
- [ ] No

ShridharGoel:
Fixes #1672 

Checklist:

Changes: 

- Proper attendee IDs are being set.
- Attendee check-in status gets updated after scanning.

### DIFF
--- a/app/src/main/java/com/eventyay/organizer/core/attendee/list/AttendeesFragment.java
+++ b/app/src/main/java/com/eventyay/organizer/core/attendee/list/AttendeesFragment.java
@@ -158,8 +158,8 @@ public class AttendeesFragment extends BaseFragment<AttendeesPresenter> implemen
         } else {
             fastItemAdapter.withComparator((Attendee a1, Attendee a2) -> a1.getFirstname().compareTo(a2.getFirstname()), true);
         }
-        fastItemAdapter.setNewList(getPresenter().getAttendees());
-        binding.setVariable(BR.attendees, getPresenter().getAttendees());
+        fastItemAdapter.setNewList(attendeeList);
+        binding.setVariable(BR.attendees, attendeeList);
         binding.executePendingBindings();
     }
 
@@ -182,7 +182,6 @@ public class AttendeesFragment extends BaseFragment<AttendeesPresenter> implemen
         setupRefreshListener();
         setupRecyclerView();
         getPresenter().attach(eventId, this);
-        binding.setAttendees(getPresenter().getAttendees());
         getPresenter().start();
     }
 
--- a/app/src/main/java/com/eventyay/organizer/core/attendee/list/AttendeesFragment.java
+++ b/app/src/main/java/com/eventyay/organizer/core/attendee/list/AttendeesFragment.java
@@ -294,8 +294,14 @@ public class AttendeesFragment extends BaseFragment<AttendeesPresenter> implemen
             refreshLayout.setRefreshing(false);
             attendeeList.clear();
             getPresenter().loadAttendeesPageWise(FIRST_PAGE, true);
-            fastItemAdapter.setNewList(attendeeList);
         });
+    }
+
+    @Override
+    public void onResume() {
+        super.onResume();
+        attendeeList.clear();
+        getPresenter().loadAttendeesPageWise(FIRST_PAGE, false);
     }
 
     // View Implementation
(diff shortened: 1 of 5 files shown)


iamareebjamal:
Please summarize the changes and explain why it wasn't working and how you fixed it

ShridharGoel:
The `getAttendeesPagewise()` method fetches the list at a particular page and appends it to the earlier list which was fetched till then. So, the `updateLocal()` method wasn't taking the correct list, it was using `attendeesList` in the presenter which had the list of attendees of just one page and not the complete list. So, I have created `getAttendeeList()` in AttendeesFragment and replaced `attendeeList` with `getView().getAttendeeList()` in AttendeesPresenter.

Apart from this, after scanning, the list was not updating until refreshed. Therefore, I've added this in the ScanQRPresenter:
```
attendee.isCheckedIn = !attendee.isCheckedIn;
        attendeeRepository.scheduleToggle(attendee)
            .subscribe(() -> {
                // Nothing to do
            }, Logger::logError);
```

ShridharGoel:
One issue that still exists: The attendee IDs do not update when the fragment is opened for the first time. After refreshing once, the correct values show. This is because the attendees are not getting stored in the local database until refreshed once. Although we are shifting to Room and that would probably fix this, but I think this needs to be fixed right now. So, I'm working on it. 
@iamareebjamal Please tell if you have any suggestions on this.

iamareebjamal:
I'm not able to understand why this part was needed

```java
attendeeRepository.scheduleToggle(attendee)
    .subscribe(() -> {
        // Nothing to do
    }, Logger::logError);
```

ShridharGoel:
> I'm not able to understand why this part was needed
> 
> ```java
> attendeeRepository.scheduleToggle(attendee)
>     .subscribe(() -> {
>         // Nothing to do
>     }, Logger::logError);
> ```

To update the attendee in the local DB?

```
@NonNull
    public Completable scheduleToggle(Attendee attendee) {
        return repository
            .update(Attendee.class, attendee)
            .concatWith(completableObserver -> {
                AttendeeCheckInWork.scheduleWork();
                if (!repository.isConnected())
                    completableObserver.onError(new Exception("No network present. Added to job queue"));
            })
            .subscribeOn(Schedulers.io());
    }
```

ShridharGoel:
`attendee.isCheckedIn = !attendee.isCheckedIn;`

This is not needed since it is being handled here:
```
if (toCheckIn) {
            getView().showMessage(
                attendee.isCheckedIn ? R.string.already_checked_in : R.string.now_checked_in,
                true
            );
            attendee.isCheckedIn = true;
        } else if (toCheckOut) {
            getView().showMessage(
                attendee.isCheckedIn ? R.string.now_checked_out : R.string.already_checked_out,
                true
            );
            attendee.isCheckedIn = false;
        }
```

iamareebjamal:
But it was working before without it. Why is it needed now?

ShridharGoel:
> But it was working before without it. Why is it needed now?

Wasn't working, the attendee status did not change until the list is refreshed.

iamareebjamal:
Again, before the pagination changes, it was working. You can check in the old APK. It means this was triggered somewhere else as well. Please check the reason why it stopped working

ShridharGoel:
Okay, checking it.

ShridharGoel:
Scanning is working as expected. Earlier, I think I was pressing the back button quickly after scanning to go back to the list so maybe that was the reason of the list not showing updated changes or maybe it could have been because of internet connectivity issue. Reverted the changes in ScanQRPresenter.

There was an issue because of which on loading second and further pages, the data of the earlier page was getting deleted from the local DB, due to which attendee IDs started showing as 0 for the earlier pages. Updated the AttendeeRepositoryImpl to fix this.

ShridharGoel:
> One issue that still exists: The attendee IDs do not update when the fragment is opened for the first time. After refreshing once, the correct values show. This is because the attendees are not getting stored in the local database until refreshed once. Although we are shifting to Room and that would probably fix this, but I think this needs to be fixed right now. So, I'm working on it.
> @iamareebjamal Please tell if you have any suggestions on this.

This is also fixed.

PR #1690, opened 2019-06-04
fix: Set latitude and longitude on event creation

### LINKED ISSUE #1689: Event map remains blank for events created through app
**Describe the bug**
Latitude and longitude details of the location entered by the user remain unavailable due to which maps are shown as blank.

- [x] Yes
- [ ] No

ShridharGoel:
Fixes #1689


app/src/main/java/com/eventyay/organizer/core/event/create/CreateEventViewModel.java
```diff
@@ -149,6 +153,15 @@ public void createEvent() {
 
         nullifyEmptyFields(event);
 
+        if (event.getLocationName() != null) {
+
+            double latitude = LocationUtils.getLatitude(OrgaProvider.context, event.getLocationName());
+            double longitude = LocationUtils.getLongitude(OrgaProvider.context, event.getLocationName());
```
iamareebjamal: Why are you using geocoder twice when you should only once

ShridharGoel:
> Use a map picker than this please

@iamareebjamal It is already there but hidden because Google Places API key needs to be added.

iamareebjamal:
OK, I'll add that. You verify that it's working on your personal jey

ShridharGoel:
@iamareebjamal The code seems good to me. I couldn't verify it since using Places API key requires to set up billing.

iamareebjamal:
It allows 1 query per day for free

ShridharGoel:
Yes, but it still requires to set up billing before using. I had generated a key and tried using it but the map opens and then closes instantly and logs show: 
`Places API for Android does not seem to be enabled for your app.`

Screenshot from 2019-06-05 00-40-18

iamareebjamal:
OK, I'll give you a key on personal chat on gitter tomorrow

ShridharGoel:
@iamareebjamal Place Picker is deprecated and cannot be used now: https://stackoverflow.com/a/55524719

Should we use Places Autocomplete instead?

mariobehling:
No need for autocompletion right now. We can come back to it later. Just lat and lon are important.

ShridharGoel:
@mariobehling #1737 fixes this. Hence, this can be closed.

PR #1694, opened 2019-06-05
feat: Implement notifications section

### LINKED ISSUE #1693: Add notifications section
**Describe the feature you'd like**
List of notifications.

- [x] Yes
- [ ] No

ShridharGoel:
Fixes #1693 

Notifications


app/src/main/java/com/eventyay/organizer/data/db/configuration/OrgaDatabase.java
```diff
@@ -310,4 +311,20 @@ public void onPreMigrate() {
         }
     }
 
+    @Migration(version = 20, database = OrgaDatabase.class)
```
liveHarshit: `version = VERSION`

ShridharGoel: Wouldn't it be better to keep it like this only since when migrating to the next version we'll have to change this again? Because older versions are not removed.

iamareebjamal: `version = VERSION` is wrong. Migrations will never run correctly if written that way

iamareebjamal:
Remove divider

ShridharGoel:
> Remove divider

Done.

PR #1699, opened 2019-06-07
chore: Add 'Use another URL' button for base URL

### LINKED ISSUE #1698: Add 'Use another URL' option instead of current checkbox system for base URL
**Describe the chore**
Currently, when the user unticks the checkbox, a field to enter another base URL comes up.

Instead of the checkbox, the default URL should be there along with an option to use another URL.

- [x] Yes
- [ ] No

ShridharGoel:
Fixes #1698 

GIF for the change:

Use another URL


ShridharGoel:
@mariobehling The production app always uses `api.eventyay.com`. These settings are only for the debug app. 

> 2. open-event-api... -> delete it. Don't show it! And use eventyay.com.

Yes, this will done in #1700. I'll open a PR by tonight.

PR #1708, opened 2019-06-11
fix: Retain 'Use picker' checkbox state on screen rotation

### LINKED ISSUE #1578: User Picker in EventDetailStepOne is not retained after rotating screen
**Describe the bug**
In EventDetailStepOne, when a "user picker" CheckBox is ticked while the other is not, rotating the screen will make both of them get ticked or not ticked together.

Besides, the screenshot below also shows that a toast message is always displayed whenever screen rotates

**Screenshots**

**Additional context**
This is probably caused because we includes the CheckBox in the view with same id so Android automatically retain the current state of one CheckBox for both of them

No

ShridharGoel:
Fixes #1578


mariobehling:
Your build is failing. Please resolve travis.

PR #1725, opened 2019-06-16
chore: Simplify the app

### LINKED ISSUE #1724: Simplify the app
**Describe the chore**
Extra features need to be commented out from the app, as discussed.

- [x] Yes
- [ ] No

ShridharGoel:
Fixes #1724

### DIFF
--- a/app/src/main/java/com/eventyay/organizer/core/main/FragmentNavigator.java
+++ b/app/src/main/java/com/eventyay/organizer/core/main/FragmentNavigator.java
@@ -69,9 +69,9 @@ class FragmentNavigator {
             case R.id.nav_dashboard:
                 fragment = EventDashboardFragment.newInstance(eventId);
                 break;
-            case R.id.nav_sell:
+            /*case R.id.nav_sell:
                 fragment = CreateOrderFragment.newInstance(eventId);
-                break;
+                break;*/
             case R.id.nav_attendees:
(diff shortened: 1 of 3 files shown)


iamareebjamal:
What was the point of implementing roles and speaker call if it was to commented out?

iamareebjamal:
Instead of commenting it out. Make it configurable through env variables in gradle file.

ShridharGoel:
@iamareebjamal In the meeting, @mariobehling told me to comment out these features as they are not needed in the app and can be used from the frontend when required.

iamareebjamal:
That'll make the efforts made into the app development futile and development hours wasted. Just add the option to enable and disable features through configuration and we'll add the necessary values in the build process

Ref: https://github.com/luggit/react-native-config/blob/master/android/dotenv.gradle

mariobehling:
@iamareebjamal We decided in the meeting to comment it out for now as we only have one developer working on this. We want to focus on the basic functionality first. We can add an "expert mode" to the app later in GSoC.

@ShridharGoel We are now approaching the second term of GSoC. You should be sticking to best practices by now. Please always add screenshots and for major changes a test apk. It should have been here. There is no description in the PR here neither. Simply saying "fixes issue" is not enough. We need a description, screenshot etc. See how other people are doing it e.g. in the frontend project. You are not following best practices.

PR #1808, opened 2019-07-12
feat: Implement sound for check-in and check-out

### LINKED ISSUE #1747: Improve "Device Settings"
* Take out option "Collect attendee information" (and ensure we are collecting attendee information by default)
* Add option "Play Sounds" (Enable sounds for user interface events) and implement different sounds on successful and unsuccessful check in

ShridharGoel:
Fixes #1747 

Changes: 

- Implemented sound for check-in and check-out and added option to toggle on/off the feature to play sounds in Device Settings.
- Removed 'Collect attendee information'.

Sound used: https://freesound.org/people/thisusernameis/sounds/426888


app/src/main/java/com/eventyay/organizer/core/attendee/list/SwipeController.java
```diff
@@ -57,6 +65,11 @@ public int getMovementFlags(RecyclerView recyclerView, RecyclerView.ViewHolder v
     public void onSwiped(RecyclerView.ViewHolder viewHolder, int direction) {
         int swipedPosition = viewHolder.getAdapterPosition();
         attendeesViewModel.toggleCheckInState(attendeeList, swipedPosition);
+
+        if (playSound) {
+            MediaPlayer mediaPlayer = MediaPlayer.create(context, R.raw.check_in_sound);
```
liveHarshit: I think raw fine not required. The media player can convert text to speech.

ShridharGoel: Text to speech? Here, we are just playing a sound when a user is checked in.

liveHarshit: Yes, for that media player can speak the same from string also. It will reduce the app size.

ShridharGoel: But how can the this type of sound be written as string? It doesn't contain words which we can convert using Text to Speech.

liveHarshit: Oh, I was thinking that it is speaking checked in.

PR #1813, opened 2019-07-15
feat: Implement Mapbox for event location in F-Droid build variant

### LINKED ISSUE #1809: Implement Mapbox for adding event location in F-Droid build variant
**Describe the feature you'd like**
Implement Mapbox for adding event location in F-Droid build variant.

- [x] Yes
- [ ] No

ShridharGoel:
Fixes #1809

GIF for the change:

Implement Mapbox for event location in F-Droid build variant


app/build.gradle
```diff
@@ -62,17 +63,19 @@ android {
             proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
 
             buildConfigField "String", "DEFAULT_BASE_URL", '"https://api.eventyay.com/v1/"'
+            buildConfigField "String", "MAPBOX_ACCESS_TOKEN", "\"${MAPBOX_ACCESS_TOKEN}\""
```
iamareebjamal: Should be with Google Places API key, not in between base URLs

iamareebjamal:
Note that app should not crash if there is no access token

app/build.gradle
ShridharGoel: Updated.

ShridharGoel:
> Note that app should not crash if there is no access token

Done.

PR #1895, opened 2019-08-16
fix: Prevent Account section's action bar title from changing

### LINKED ISSUE #1892: Action Bar title changes when clicking on email
**Describe the bug**
When clicking on email to redirect to mailing app, ActionBar title change from "<Account Name>" to "OrganizerDetailActivity"
Besides, it is also illogical that the person who owning the email want to send an email to himself/herself

**Screenshots**
ezgif-2-49d310e68ead

Yes

ShridharGoel:
Fixes #1892

### DIFF
--- a/app/src/main/java/com/eventyay/organizer/core/organizer/detail/OrganizerDetailFragment.java
+++ b/app/src/main/java/com/eventyay/organizer/core/organizer/detail/OrganizerDetailFragment.java
@@ -1,20 +1,12 @@
 package com.eventyay.organizer.core.organizer.detail;
 
-import androidx.annotation.Nullable;
-import androidx.lifecycle.ViewModelProvider;
-import androidx.lifecycle.ViewModelProviders;
 import android.content.Intent;
 import android.content.pm.ApplicationInfo;
 import android.content.pm.PackageManager;
-import androidx.databinding.DataBindingUtil;
 import android.graphics.Bitmap;
 import android.graphics.BitmapFactory;
 import android.net.Uri;
 import android.os.Bundle;
-import com.google.android.material.snackbar.Snackbar;
-import androidx.swiperefreshlayout.widget.SwipeRefreshLayout;
-import androidx.appcompat.app.ActionBar;
-import androidx.appcompat.app.AppCompatActivity;
 import android.view.LayoutInflater;
 import android.view.Menu;
 import android.view.MenuInflater;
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/eventyay/organizer/core/organizer/detail/OrganizerDetailFragment.java
```diff
@@ -185,6 +186,25 @@ public void onStart() {
...
+            String firstName = user.getFirstName();
+            String lastName = user.getLastName();
+            String email = user.getEmail();
+
+            if (firstName != null)
+                ((AppCompatActivity) getActivity()).getSupportActionBar().setTitle(
+                    firstName + " " + lastName);
+            else
+                ((AppCompatActivity) getActivity()).getSupportActionBar().setTitle(email);
+        }
```
iamareebjamal: Where was it previously being set?

iamareebjamal:
@anhanh11001 That's set in AndroidManifest

app/src/main/java/com/eventyay/organizer/core/organizer/detail/OrganizerDetailFragment.java
ShridharGoel: In the XML file.

iamareebjamal: Why wasn't it working?

ShridharGoel: Whenever the fragment was being resumed, the title was getting changed to `OrganizerDetailActivity` because of the `getTitle()` method of the BaseFragment.

ShridharGoel: The `onResume()` method in the BaseFragment:

```
@Override
public void onResume() {
    super.onResume();
    setTitle(getString(getTitle()));
}
```

iamareebjamal: So why not override `getTitle`?

ShridharGoel: `getTitle()` is already overridden and `OrganizerDetailActivity` is being set from there. Is it possible to set the organizer name from there since it takes int resource value and we cannot change values in strings.xml programatically at runtime?

iamareebjamal: Got it

app/src/main/java/com/eventyay/organizer/core/organizer/detail/OrganizerDetailFragment.java
```diff
@@ -185,6 +186,25 @@ public void onStart() {
...
+        if (user != null) {
+            String firstName = user.getFirstName();
+            String lastName = user.getLastName();
+            String email = user.getEmail();
+
+            if (firstName != null)
+                ((AppCompatActivity) getActivity()).getSupportActionBar().setTitle(
+                    firstName + " " + lastName);
+            else
+                ((AppCompatActivity) getActivity()).getSupportActionBar().setTitle(email);
```
iamareebjamal: Eliminate duplication `((AppCompatActivity) getActivity()).getSupportActionBar()`

ShridharGoel: Updated.

PR #1959, opened 2019-10-08
fix: last name shows as "null"

### LINKED ISSUE #1939: Last name shows as "null"
**At the time of signup,Last name field is optional,so if we don't fill this field then it shows last name as null**

**Screenshots**

**Smartphone Info:**

|               | |
|---------------|-|
|Device         ||
|Android Version 6.0|    |

**Additional context**

- [-] Yes
- [ ] No
- Other:

thatfedupguy:
Fixes #1939 

Changes: At the time of signup, the Last name field is optional, so if we don't fill this field then it shows the last name as null in navigation drawer as there was no check on the last name to be null. I have added a check on the user's last name to be "" if null otherwise the last name entered by the user.

Screenshots for the change:
Screenshot_20191009-011540

### DIFF
--- a/app/src/main/res/layout/main_nav_header.xml
+++ b/app/src/main/res/layout/main_nav_header.xml
@@ -55,7 +55,7 @@
                 <TextView
                     android:layout_width="wrap_content"
                     android:layout_height="wrap_content"
-                    android:text='@{ user.firstName == null ? user.email : user.firstName + " " + user.lastName }'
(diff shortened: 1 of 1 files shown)


ShridharGoel (approved):

iamareebjamal (changes requested):
Change PR title according to the guidelines

thatfedupguy:
@iamareebjamal  I have changed the PR title accordingly.

iamareebjamal (approved):

PR #2008, opened 2019-12-18
fix: Multiple Screens removed 

### LINKED ISSUE #2007: Multiple CreateEventActivities open on multiple taps
**Describe the bug**
Multiple CreateEventActivity open up when the add('+') is pressed multiple times simultaneously.
Similar bug is found on tapping the profile icon(found in navigation drawer) multiple times.

**To Reproduce**
1. Go to the App
2. Tap on add Event button or the profile icon multiple times.
3. Now, as you press the backPress button, you land up in CreateEvent Activity or Profile Activity
 respectively depending on what you tap.

**Expected behavior**
You should have reached the home screen instead of the CreateEvent Activity.

**Logs**

**Screenshots**

**Smartphone Info:**

|               | |
|---------------|-|
|Device         |Samsung Galaxy J8|
|Android Version|Pie 9.0|

**Additional context**

- [X] Yes
- [ ] No
- Other:

Kalit31:
Fixes #2007 

Changes: Added a flag in EventListFragment and MainActivity to prevent opening of multiple screens on button tap.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -220,7 +220,7 @@ dependencies {
     implementation 'com.f2prateek.rx.preferences2:rx-preferences:2.0.0'
 
     // Fast Adapter
-    implementation "com.mikepenz:fastadapter:4.1.1"
+    implementation "com.mikepenz:fastadapter:4.1.2"
     implementation 'com.timehop.stickyheadersrecyclerview:library:0.4.3'
     implementation 'me.xdrop:fuzzywuzzy:1.1.9'
 


adityastic (commented):
Hi Kalit, thanks for dropping a pull request. What I suggest is you should attach the onclick listeners onResume and set them null on onDetach and when the button is clicked. This helps you avoid memory leaks and clears situations like these.

adityastic:
Edit: It would be a better approach to just control this field from the viewmodel. That would avoid the state changes as well.

iamareebjamal:
Not sure this is something we should fix. This will needlessly add code to the app when not needed. It's a bug (if even considered one) in the Android Framework, not in the app

PR #2016, opened 2019-12-30
Added Dutch translation

Vistaus:
:-)

### DIFF
--- a/app/src/main/res/values-nl/strings.xml
+++ b/app/src/main/res/values-nl/strings.xml
@@ -1,0 +1,1029 @@
+<resources>
+    <string name="checked_in">Ingecheckt</string>
+    <string name="checked_out">Uitgecheckt</string>
+    <string name="check_in">Inchecken</string>
+    <string name="event_title">Naam van evenement</string>
+    <string name="tickets_sold">Aantal verkochte tickets</string>
+    <string name="to"> Aan </string>
+    <string name="comma"> , </string>
+    <string name="divide">/</string>
+    <string name="attendance">Aanwezigheid</string>
+    <string name="scan_qr_code">QR-code scannen</string>
+    <string name="first_name">Voornaam (optioneel)</string>
+    <string name="last_name">Achternaam (optioneel)</string>
+    <string name="email">E-mailadres</string>
+    <string name="edit_profile">Profiel aanpassen</string>
+    <string name="update">Bijwerken</string>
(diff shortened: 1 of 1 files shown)


app/src/main/res/values-nl/strings.xml
```diff
@@ -0,0 +1,1029 @@
...
+    <string name="developers">Ontwikkelaars</string>
+    <string name="github_repo_link">https://github.com/fossAzië/open-event-organizer-android</string>
+    <string name="github_developers_link">https://github.com/fossAzië/open-event-organizer-android/graphs/contributors</string>
+    <string name="third_party_licenses">Gebruikte externe bibliotheken</string>
+
+    <string-array name="timezones">
+        <item>Afrika/Abidjan</item>
+        <item>Afrika/Accra</item>
+        <item>Afrika/Addis_Abbeba</item>
+        <item>Afrika/Algerije</item>
```
iamareebjamal: This should not be translated

Vistaus: Why not? Quite a few of them have different names in Dutch, so...

iamareebjamal: Because they're standard timezone names regardless of the language

Vistaus: Hmm, with other projects, this is not an issue. But fair enough, I'll change it back then :)

Vistaus: Done!

iamareebjamal: Other projects like? AFAIK, you can't just send any translation of a timezone name to a date API, it won't understand it. I wished this was in code and not in strings or at least translatable="false" was set on it

iamareebjamal: BRW, the build is failing after the addition of changes

Vistaus: Why is it failing still?

iamareebjamal: You can know by clicking on details

Vistaus: It doesn't tell me anything useful. Seems like a few tests and stuff are failing, so that's something for the dev team to look at.

iamareebjamal: Not really, build fails after these changes

Vistaus: But why? I changed the timezone list back...

